PR #10, opened 2024-04-26
GREEN-40 relocate the /removed endpoint behind debug middleware

tonititi:
GREEN-40 relocated the /removed endpoint behind debug middleware in the src/API.ts file


jairajdev (changes requested):
Plz, check the comment on the task. thanks
https://linear.app/shm/issue/GREEN-40/put-archiver-debug-endpoints-behind-debug-middleware

PR #44, opened 2024-06-27
BLUE-134

jairajdev:
Port the logic from shardus core to modify the network account itself on the change cycle


src/GlobalAccount.ts
```diff
@@ -35,6 +37,149 @@ export function setGlobalNetworkAccount(account: AccountDB.AccountCopy): void {
...
+  const activeConfigChanges = new Set<string>()
+  for (const change of changes) {
+    // skip future changes
+    if (change.cycle > cycleNumber) {
+      continue
+    }
+    const changeHash = Crypto.hashObj(change)
+    // skip handled changes
+    if (appliedConfigChanges.has(changeHash)) {
+      activeConfigChanges.add(changeHash)
```
urnotsam, line 62: Mentioned in core PR as well.. Should we be adding activeConfigChanges here.  The comment implied to skip handled changes

scripts/update_config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+crypto.init(config.ARCHIVER_HASH_KEY)
+
+const DEV_KEYS = {
+  pk: config.ARCHIVER_PUBLIC_KEY,
+  sk: config.ARCHIVER_SECRET_KEY,
+}
+
+function sign<T>(obj: T, sk: string, pk: string): T & any {
+  const objCopy = JSON.parse(crypto.stringify(obj))
```
urnotsam: Should this JSON.parse be using the new safe safeJsonParse?

PR #52, opened 2024-07-19
Sys 188 add GitHub ci

TieDyedSheep:
DO NOT MERGE without approval from Andrew!

The main ci.yml for this was already migrated, this makes some minor tweaks to fix an AI PR review agent, renames some of the jobs so their purpose is more clear and corrects a bug in the quoting for the linting command. This should have no functional impact to the repository.


TieDyedSheep:
There were prettier fixes that failed the build but it only impacted 3 files and they were fairly minor so i ran them and pushed the changes. Take a look and let me know if that's cool or if I should revert that commit.

PR #55, opened 2024-07-31
BLUE-191 Fixed devPubKeys update and added to use maxCyclesShardDataToKeep config from the network

jairajdev:


afostr:
this was merged to dev , closing pr

PR #56, opened 2024-07-31
Beta1.12.1rc0  security review archive server

afostr:


src/GlobalAccount.ts
```diff
@@ -35,6 +37,145 @@ export function setGlobalNetworkAccount(account: AccountDB.AccountCopy): void {
...
+
+    let appConfigs = []
+    if (thisChange.appData) {
+      appConfigs = generatePathKeys(thisChange.appData, 'appdata.')
+    }
+    const shardusConfigs: string[] = generatePathKeys(thisChange.change)
+
+    const allConfigs = appConfigs.concat(shardusConfigs)
+
+    for (const config of allConfigs) {
```
mssabr01, line 141: Perhaps I am misunderstanding this. If a specific config path has been changed enough times in the same call of this function, then any further changes to this path are discarded? Will this discard valid changes, or will it just put them off until the next cycle? Is it possible for a valid change to be delayed long enough to be invalid via configChangeMaxCyclesToKeep?

afostr: I think this was cut an pasted from the validator o that the archive server can understand these values.   
the goal of prune work changes is to keep the last value when there have been past changes.
i.e  what if we change some config daily reward rates.  this will prune out the old change history for that config

afostr: I will ask Red to help double check where the code was transplanted

PR #62, opened 2024-08-12
Add version check for firstnode in restore network

tanuj-shardeum:


src/API.ts
```diff
@@ -98,16 +114,51 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+      
+
+      const networkAccount: AccountDB.AccountsCopy | string = getGlobalNetworkAccount(false)
+      Logger.mainLogger.debug('networkAccount', networkAccount)
+
+      if (networkAccount) {
+        if (
+          typeof networkAccount != 'string' &&
+          !NodeList.isValidVersion(
+            networkAccount?.data?.current?.minVersion,
```
thantsintoe, line 145: How will the archiver knows the data type and format of the global/network account from the Shardeum? For instance, if archiver is used other app layers (other than Shardeum) that use different format for network account, this will fail.

src/API.ts
```diff
@@ -98,16 +114,51 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+        Logger.mainLogger.debug('POST /nodelist firstNode:', firstNode)
+        Logger.mainLogger.debug('POST /nodelist appJoinData:', appJoinData)
+      }
+      
+
+      const networkAccount: AccountDB.AccountsCopy | string = getGlobalNetworkAccount(false)
+      Logger.mainLogger.debug('networkAccount', networkAccount)
+
+      if (networkAccount) {
+        if (
```
thantsintoe, line 142: It would be good if this `if` check can be simplified for better readability

thantsintoe (changes requested):

PR #67, opened 2024-08-23
Add txDigest cron and api server initial implementation

tanuj-shardeum:


src/Data/Collector.ts
```diff
@@ -933,8 +935,17 @@ export const storeReceiptData = async (
...
+    const processedTx: ProcessedTransaction.ProcessedTransaction = {
+      txId: tx.txId,
+      cycle: cycle,
+      txTimestamp: tx.timestamp,
+      txApplyTimestamp: null,
+    }
+
     // await Transaction.insertTransaction(txObj)
     combineTransactions.push(txObj)
+    combineProcessedTxs.push(processedTx)
```
aniketdivekar: I think I asked this question in call already, but at this point, we know that processedTx has the transaction receipt, correct?

src/dbstore/index.ts
```diff
@@ -57,6 +58,16 @@ export const initializeDB = async (config: Config): Promise<void> => {
     originalTxDataDatabase,
     'CREATE INDEX if not exists `originalTxsData_txId_idx` ON `originalTxsData` (`txId`)'
   )
+
+  // Transaction digester service tables
+  await runCreate(
+    processedTxDatabase,
+    'CREATE TABLE if not exists `processedTxs` (`txId` TEXT NOT NULL, `cycle` BIGINT NOT NULL, `txTimestamp` BIGINT NOT NULL, `txApplyTimestamp` BIGINT, PRIMARY KEY (`txId`))'
```
aniketdivekar: can we use VARCHAR instead of TEXT?

src/dbstore/processedTxs.ts
```diff
@@ -0,0 +1,102 @@
...
+  txTimestamp: number
+  txApplyTimestamp: number
+}
+
+export async function insertProcessedTx(processedTx: ProcessedTransaction): Promise<void> {
+  try {
+    const fields = Object.keys(processedTx).join(', ')
+    const placeholders = Object.keys(processedTx).fill('?').join(', ')
+    const values = extractValues(processedTx)
+    const sql = 'INSERT OR REPLACE INTO processedTxs (' + fields + ') VALUES (' + placeholders + ')'
```
aniketdivekar: We should use ON CONFLICT clause instead of INSERT OR REPLACE.

src/dbstore/processedTxs.ts
```diff
@@ -0,0 +1,102 @@
...
+    )
+  }
+}
+
+export async function bulkInsertProcessedTxs(processedTxs: ProcessedTransaction[]): Promise<void> {
+  try {
+    const fields = Object.keys(processedTxs[0]).join(', ')
+    const placeholders = Object.keys(processedTxs[0]).fill('?').join(', ')
+    const values = extractValuesFromArray(processedTxs)
+    let sql = 'INSERT OR REPLACE INTO processedTxs (' + fields + ') VALUES (' + placeholders + ')'
```
aniketdivekar: same as above.

src/dbstore/processedTxs.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    if (!txIds) {
+      return []
+    }
+
+    txIds.sort()
+    return txIds
+  } catch (e) {
+    Logger.mainLogger.error(e)
```
aniketdivekar: can you add some identifier text here? something like Logger.mainLogger.error('error in querySortedTxsBetweenCycleRange', e)

src/txDigestAPIserver.ts
```diff
@@ -0,0 +1,90 @@
...
+    return StringUtils.safeStringify(payload)
+  })
+
+  // Register API routes
+  registerRoutes(server as FastifyInstance<Server, IncomingMessage, ServerResponse>)
+
+  // Start server and bind to port on all interfaces
+  server.listen(
+    {
+      port: 8084,
```
aniketdivekar: TODO: make it configurable

src/txDigestAPIserver.ts
```diff
@@ -0,0 +1,90 @@
...
+      host: '0.0.0.0',
+    },
+    (err) => {
+      Logger.mainLogger.debug('Listening', config.ARCHIVER_PORT)
```
aniketdivekar: this needs a fix in `TXDigestAPI Listening' and respective port

src/txDigestAPIserver.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (err) {
+        server.log.error(err)
+        process.exit(1)
+      }
+      Logger.mainLogger.debug('Archive-server has started.')
```
aniketdivekar: please change the log message

src/txDigester.ts
```diff
@@ -0,0 +1,70 @@
...
+    startSaving(join(baseDir, logsConfig.dir))
+  }
+
+  await dbstore.initializeDB(config)
+
+  await txDigesterDB.initializeDB(config)
+
+  const ARCHIVER_STATUS_CHECK_URL = `http://${config.ARCHIVER_IP}:${config.ARCHIVER_PORT}/status`
+
+  cron.schedule('*/5 * * * *', async () => {
```
aniketdivekar: TODO: take from config

src/txDigester/api.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+}>
+
+/* To-Do: Add LRU cache for the tx-digests */
+export function registerRoutes(server: FastifyInstance<Server, IncomingMessage, ServerResponse>): void {
+  server.get('/api/tx-digests', async (_request: GetTxDigestsRequest, reply) => {
+    const cycleStart = Number(_request.query?.cycleStart)
+    const cycleEnd = Number(_request.query?.cycleEnd)
+
+    if (isNaN(cycleStart) || isNaN(cycleEnd) || cycleEnd <= cycleStart || cycleStart < 0 || cycleEnd < 0) {
```
aniketdivekar: Let's add one more check, diff between cycleEnd and cycleStart should be <=10000 (so we fetch max 1000 records in a call assuming cycleDiff will never be less than 10).

src/txDigester/txDigestFunctions.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export const processAndInsertTxDigests = async (
+  lastCheckedCycle: number,
+  latestCycleCounter: number
+): Promise<void> => {
+  console.log('Processing and inserting txDigests from cycle: ', lastCheckedCycle, ' to ', latestCycleCounter)
+  const batchSize = config.txDigest.cycleDiff
+  let currentCycle = lastCheckedCycle
+  let endCycle = currentCycle + batchSize - 1
```
aniketdivekar: math works well here right? like 0-9, 10-11 etc?

src/txDigester/txDigestFunctions.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export const getTxIds = async (startCycle: number, endCycle: number): Promise<string[]> => {
+  const sortedTxIds = await processedTxs.querySortedTxsBetweenCycleRange(startCycle, endCycle)
+  return sortedTxIds
+}
+
+export const getHash = async (cycle: number): Promise<string> => {
+  const txDigestHash = await txDigest.queryByEndCycle(cycle)
+  return txDigestHash ? txDigestHash.hash : '0x0'
```
aniketdivekar: we should have some checks here. Like return '0x0' only if cycle = 0. If not, then throw error.

src/txDigester/txDigests.ts
```diff
@@ -0,0 +1,76 @@
...
+  txCount: number
+  hash: string
+}
+
+export async function insertTransactionDigest(txDigest: TransactionDigest): Promise<void> {
+  try {
+    const fields = Object.keys(txDigest).join(', ')
+    const placeholders = Object.keys(txDigest).fill('?').join(', ')
+    const values = extractValues(txDigest)
+    const sql = 'INSERT OR REPLACE INTO txDigests (' + fields + ') VALUES (' + placeholders + ')'
```
aniketdivekar: same suggestion as above for using `ON CONFLICT` clause

src/txDigester/txDigests.ts
```diff
@@ -0,0 +1,76 @@
...
+    return lastProcessedDigest
+  } catch (e) {
+    console.error(e)
+    return null
+  }
+}
+
+export async function queryByEndCycle(endCycle: number): Promise<TransactionDigest> {
+  try {
+    const sql = `SELECT * FROM txDigests WHERE cycleEnd=? LIMIT 1`
```
aniketdivekar: do we have a separate index on cycleEnd? if not, we should add one. I see it in where clause as well as order by clause.

src/Data/Collector.ts
tanuj-shardeum: Yes

src/dbstore/index.ts
tanuj-shardeum: Done

src/dbstore/processedTxs.ts
tanuj-shardeum: Done

src/dbstore/processedTxs.ts
tanuj-shardeum: Done

src/txDigester/api.ts
tanuj-shardeum: Done

src/txDigester/txDigestFunctions.ts
tanuj-shardeum: Yes

src/txDigester/txDigestFunctions.ts
tanuj-shardeum: Done

src/txDigester/txDigests.ts
tanuj-shardeum: Done

src/txDigester/txDigests.ts
tanuj-shardeum: I guess instead of having cycleStart and cycleEnd as a composite primary key, we can just use cycleEnd as a primary key. Added that

src/dbstore/processedTxs.ts
tanuj-shardeum: Done

src/txDigestAPIserver.ts
tanuj-shardeum: Done

src/txDigestAPIserver.ts
tanuj-shardeum: Done

src/txDigestAPIserver.ts
tanuj-shardeum: Done

src/txDigester.ts
tanuj-shardeum: Done

src/txDigester/txDigests.ts
aniketdivekar: This way we can have records like 10-19, 10-29. I guess we should just switch the order of cycleStart and cycleEnd in composite primary key and we should get the advantage of uniqueness of the range as well as improved performance.

jairajdev (dismissed):

jairajdev (dismissed):

jairajdev (dismissed):

jairajdev (dismissed):

PR #76, opened 2024-09-04
BLUE-271 fix: applyTimestamp added to processed tx

achal-singh:
### Summary
**`applyTimestamp`** was not set correctly in case of processed transaction logic, this PR fixes that.

### DIFF
--- a/src/Data/Collector.ts
+++ b/src/Data/Collector.ts
@@ -968,7 +968,7 @@ export const storeReceiptData = async (
       txId: tx.txId,
       cycle: cycle,
       txTimestamp: tx.timestamp,
-      txApplyTimestamp: null,
+      applyTimestamp,
     }
 
     // await Transaction.insertTransaction(txObj)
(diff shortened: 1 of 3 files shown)


jairajdev (approved):

src/Data/Collector.ts
```diff
@@ -1157,7 +1157,7 @@ export const storeAccountData = async (restoreData: StoreAccountParam = {}): Pro
         txId: receipt.data.txId || receipt.txId,
         cycle: receipt.cycleNumber,
         txTimestamp: receipt.timestamp,
-        txApplyTimestamp: null,
+        applyTimestamp: receipt.timestamp,
```
devendra-shardeum, line 1160: @jairajdev  what's the difference b/w applyTimestamp and txTimestamp ? both are receiving the same value ..i.e. receipt.timestamp and also what's the reasoning behind renaming applyTimestamp from txApplyTimestamp? I hope it won't break the other flow! in archive-server

mssabr01 (approved):

src/Data/Collector.ts
jairajdev: txTimestamp is the timestamp set before putting the tx in the tx queue
txApplyTimestamp is the timestamp when the tx is processed/consensused

but in this `storeAccountData` function which was used to restore the tx receipt from the first node in the network ( this feature is not used anymore ), since there is no way we can calculate the txApplyTimestamp, we have to set the txTimestamp.

devendra-shardeum: but what's the need to have 2 fields receiving the same value.. if applyTimestamp value can't be correctly calculated and inserted, why shouldn't it be removed ? what's the purpose of having this extra field then ? @jairajdev

PR #80, opened 2024-09-11
BLUE-130 Global txreceipt verification

jairajdev:


src/Data/Collector.ts
```diff
@@ -432,6 +471,76 @@ export const verifyReceiptData = async (
...
   // Determine the home partition index of the primary account (executionShardKey)
   const { homePartition } = ShardFunction.addressToPartition(cycleShardData.shardGlobals, executionShardKey)
+  if (globalModification) {
+    const appliedReceipt = receipt.signedReceipt as P2PTypes.GlobalAccountsTypes.GlobalTxReceipt
+    if (config.skipGlobalTxReceiptVerification) return { success: true }
+    else {
+      const { signs } = appliedReceipt
+      // Refer to https://github.com/shardeum/shardus-core/blob/7d8877b7e1a5b18140f898a64b932182d8a35298/src/p2p/GlobalAccounts.ts#L397
+      let votingGroupCount = cycleShardData.shardGlobals.nodesPerConsenusGroup
+      if (votingGroupCount > cycleShardData.nodes.length) {
```
arhamj, line 481: Should we have a lower bound to the acceptable voting group count?

src/Data/Collector.ts
```diff
@@ -432,6 +471,76 @@ export const verifyReceiptData = async (
...
+        votingGroupCount = cycleShardData.nodes.length
+      }
+      let isReceiptMajority = (signs.length / votingGroupCount) * 100 >= 60
```
arhamj: Can we make the percentage for majority configurable?

src/Data/Collector.ts
```diff
@@ -432,6 +471,76 @@ export const verifyReceiptData = async (
...
+            'Invalid_receipt_globalModification_signs_count_less_than_60%'
+          )
+        return result
+      }
+      // Using a set to store the unique signers to avoid duplicates
+      const uniqueSigners = new Set()
+      for (const sign of signs) {
+        const { owner: nodePubKey } = sign
+        // Get the node id from the public key
+        const node = cycleShardData.nodes.find((node) => node.publicKey === nodePubKey)
```
arhamj: Is there a way to avoid this find operation on every iteration? This loop feels suboptimal.

src/Data/Collector.ts
```diff
@@ -432,6 +471,76 @@ export const verifyReceiptData = async (
...
+          if (nestedCountersInstance)
+            nestedCountersInstance.countEvent(
+              'receipt',
+              'globalModification_sign_node_not_in_execution_group_of_tx'
+            )
+          continue
+        }
+        uniqueSigners.add(nodePubKey)
+      }
+      isReceiptMajority = (uniqueSigners.size / votingGroupCount) * 100 >= 60
```
arhamj: Move to majority number to config here too.

src/Data/Collector.ts
```diff
@@ -612,10 +722,34 @@ const verifyAppliedReceiptSignatures = (
...
+  const { globalModification } = receipt
   const { txId: txid } = receipt.tx
+  if (globalModification) {
+    const appliedReceipt = receipt.signedReceipt as P2PTypes.GlobalAccountsTypes.GlobalTxReceipt
+    // Refer to https://github.com/shardeum/shardus-core/blob/7d8877b7e1a5b18140f898a64b932182d8a35298/src/p2p/GlobalAccounts.ts#L294
+
+    const { signs, tx } = appliedReceipt
+    // Using a map to store the good signatures to avoid duplicates
+    const goodSignatures = new Map()
+    for (const sign of signs) {
```
arhamj: Do we not need to exclude signatures from nodes not part of the consensus group here like on the previous file? Or are we filtering the signatures earlier and they are not reaching here if they are from invalid nodes?

src/shardeum/verifyGlobalTxReceipt.ts
```diff
@@ -0,0 +1,121 @@
...
+  Stake = 6,
+  Unstake = 7,
+  InitRewardTimes = 8,
+  ClaimReward = 9,
+  ChangeNetworkParam = 10,
+  ApplyNetworkParam = 11,
+  Penalty = 12,
+}
+
+export const verifyGlobalTxAccountChange = (
```
arhamj, line 34: Can we write unit tests for this function?

src/Data/Collector.ts
arhamj: 100 - consensus group size
60 - required

Not part of the node list: 10
Not part of the partition group: 5
85>60

sign verify
85>60

src/Data/Collector.ts
S0naliThakur: addressed

src/Data/Collector.ts
S0naliThakur: config `requiredMajorityVotesPercentage` introduced for the same

src/Data/Collector.ts
S0naliThakur: using Map now for reduced complexity

src/Data/Collector.ts

src/Data/Collector.ts
S0naliThakur: The signatures were not getting filtered earlier, only the required no of signatures was being calculated based on the valid node count. Have introduced the filtering in this section now.

src/Data/Collector.ts
```diff
@@ -432,6 +471,93 @@ export const verifyReceiptData = async (
...
+        )
+        if (nestedCountersInstance)
+          nestedCountersInstance.countEvent(
+            'receipt',
+            `Invalid_receipt_globalModification_signs_count_less_than_${config.requiredMajorityVotesPercentage}%`
+          )
+        return result
+      }
+      
+      const nodeMap = new Map<string, P2PTypes.NodeListTypes.Node>()
```
arhamj, line 505: Since this map is only going to change once per cycle we can move this up the level for further optimisation. But for now LGTM.

arhamj (dismissed):

arhamj (dismissed):

arhamj (approved):

PR #84, opened 2024-09-18
BLUE-273 Worker processed debug log flag

jairajdev:


achal-singh (dismissed):

devendra-shardeum (changes requested):
I think src/worker-process/index.ts should also have the flags around the ( console.log / console.error .. etc. ) .. right now i can see that only being done for src/primary-process/index.ts

mhanson-github:
> I think src/worker-process/index.ts should also have the flags around the ( console.log / console.error .. etc. ) .. right now i can see that only being done for src/primary-process/index.ts

As the other file was out of scope for what Jai thought we needed, I will in principle agree with you, but I think we should create a separate ticket for the additional flag wrapping.  Will have a task created

devendra-shardeum (approved):
marking it as approved since the requested changes moved to a new ticket.. https://linear.app/shm/issue/BLUE-275/additional-flag-wrapping-in-archiver-logs

PR #85, opened 2024-09-19
BLUE-276 Restrict first node selection based on the specified node ip and port

jairajdev:


src/Config.ts
```diff
@@ -184,6 +189,11 @@ let config: Config = {
     txCronSchedule: '*/5 * * * *',
   },
   workerProcessesDebugLog: false,
+  restrictFirstNodeSelection: true,
+  firstNodeInfo: {
+    IP: '127.0.0.1',
+    PORT: 4000,
```
urnotsam: shouldnt the default port be 9001? 4000 is the port of the archiver

src/API.ts
```diff
@@ -93,16 +94,22 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
         reply.send({ success: false, error: 'Signature verification failed' })
         return
       }
+      const ip = signedFirstNodeInfo.nodeInfo.externalIp
```
kun6fup4nd4, line 126: what if ip in payload will be different from actual ip of the sender ?

src/Config.ts
jairajdev: @urnotsam good catch on that. changed it to 9001 now.

urnotsam (dismissed):

urnotsam (dismissed):

kun6fup4nd4 (dismissed):

urnotsam (dismissed):

mhanson-github (approved):
I approve my own .patch file changes.

chrypnotoad:
@CodiumAI-Agent /describe

src/Config.ts
```diff
@@ -184,6 +186,8 @@ let config: Config = {
     txCronSchedule: '*/5 * * * *',
   },
   workerProcessesDebugLog: false,
+  restrictFirstNodeSelectionByPublicKey: false,
+  firstNodePublicKey: '',
```
chrypnotoad, line 190: Should this be an environment variable? When and how does this get set?

src/API.ts
chrypnotoad: The IP and Port are not even being checked in here are they? The title of this PR says restricted based off IP and Port, but I only see the public key being checked.
@mhanson-github @jairajdev

PR #86, opened 2024-09-24
BLUE-280: Add Publisher for Transaction Digest to IPFS via Web3.Storage

achal-singh:
### Summary
- The **[Transaction DIgest Cron Server](https://github.com/shardeum/archive-server/blob/dev/package.json#L19)** in the Archiver creates Transaction Digests every 10 cycles.
- This PR adds an **`ipfsPublisher`** (as a part of the Transaction DIgest Cron Server) to upload/publish the same Transaction Digest to the IPFS Network via the **[Web3.Storage](https://web3.storage/)** provider.

### Merging Notes
It'd be good to merge **[BLUE-283's PR](https://github.com/shardeum/archive-server/pull/89)** into BLUE-280 PR first before merging **`BLUE-280`** to **`dev`**.

#### Testing Notes
1. Before starting with the tests, make sure you register on the web3.storage portal with your email address and subscribe to their free-tier plan (that gives 5 GB of free storage).
2. Access your account on https://console.web3.storage and create a **`Space`**, set any name as per your preference. At the top of the screen you'll see a Root DID string (starts with **`did:key:...`**) of your Space.
3. Copy and paste the same email ID and Root DID of your space to the **[`root_did` and `admin_email` fields in the archiver-config.json](https://github.com/shardeum/archive-server/blob/dev/archiver-config.json#L63-L64)** file respectively and set the **`enableSavingToWeb3Storage`** flag to **`true`**.
4. Run **`npm run txDigestCronServer`** OR **`pm2 start --name txDigestCron npm -- run txDigestCronServer`** to launch the **`txDigestCronServer`** process.


package.json
```diff
@@ -81,7 +82,10 @@
     "@shardus/archiver-discovery": "^1.1.0",
     "@shardus/crypto-utils": "4.1.4",
     "@shardus/types": "1.2.21-1",
+    "@web3-storage/did-mailto": "^2.1.0",
```
aniketdivekar: let's use fixed versions
```suggestion
    "@web3-storage/did-mailto": "2.1.0",
```

src/Config.ts
```diff
@@ -183,7 +188,12 @@ let config: Config = {
     cycleDiff: 10,
     syncDelay: 20,
     apiServerPort: 8084,
-    txCronSchedule: '*/5 * * * *',
+    txCronSchedule: '* * * * *',
```
aniketdivekar: this line got committed by mistake with schedule change also I guess.. please revert to old schedule

src/Config.ts
```diff
@@ -94,6 +94,11 @@ export interface Config {
     syncDelay: number
     apiServerPort: number
     txCronSchedule: string
+    web3Storage: {
+      root_did: string
```
aniketdivekar: other configs are defined w/ lowerCamelCase. can we define these too using lowerCamelCase?

src/txDigester/ipfsPublisher.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Checks if the email has subscribed to a plan on the Web3.Storage Portal
+ * We'll most likely be Subscribed to the Starter/Free Plan (5GB of storage)
+ */
+const isSubscriptionActive = async (account: any): Promise<boolean> => {
+  try {
+    const plan = await account.plan.get()
+    if (!plan.ok) {
+      console.error(`❌ ${admin_email} does not have an active plan subscribed on Web3.Storage.`)
```
aniketdivekar: we should not log email id. let's mask it using industry standard logic or remove it.

src/txDigester/txDigestFunctions.ts
```diff
@@ -80,6 +81,9 @@ export const processAndInsertTxDigests = async (
 
     try {
       txDigest.insertTransactionDigest(txDigestObj)
+      if (config.txDigest.enableSavingToWeb3Storage) {
+        await uploadDigestToIPFS(txDigestObj)
```
aniketdivekar, line 85: `txDigest.insertTransactionDigest` has upsert(insert or update) like functionality if I am not wrong. What happens if we upload same cycle range to web3 storage? Will we write new data or replace existing data?
Also, I am now wondering how we will be using this storage.. who is going to access, what parameters will they use to access the data.. is there mapping between txDigest record & uploaded blob id?

src/txDigester/ipfsPublisher.ts
```diff
@@ -0,0 +1,124 @@
...
+    console.log(
+      `✅ Uploaded to IPFS Successfully for Cycles (${data.cycleStart} to ${data.cycleEnd}) @ https://${cid}.ipfs.w3s.link`
+    )
+    lastUploadTime = Date.now()
+    const storageUsed = await client.currentSpace()?.usage.get()
+    console.log(`${Number(storageUsed!.ok)} bytes used on Web3.Storage.`)
+    isPublisherActive = false
+  } catch (error) {
+    isPublisherActive = false
+    console.error(`❌ Error while Uploading Digest (w/ Hash: ${data.hash}) to IPFS: `)
```
aniketdivekar: lets log cycle range also

src/txDigester/ipfsPublisher.ts
```diff
@@ -0,0 +1,124 @@
...
+    console.error(error)
+    return
+  }
+}
+
+const isUploadPossible = (): boolean => {
```
aniketdivekar, line 65: I guess this approach may cause issue by looking at the code. Please correct me if I am wrong.

Assuming this is done to have upload functionality run exclusively. Let's say we are uploading blob for cycle 100-109. So, we block the processing by setting `isPublisherActive` flag. Since we call this method when we save the data in DB, ideally race condition shouldn't occur. But if there are simultaneous invocations, it will block upload for all other ranges who don't win the flag.

The other thing is, do we have any retry mechanism for this? What if saving in DB goes through, but upload doesn't happen? Are we able to upload it again? If not, we should add such process. Let that process take care of uploading blobs instead of uploading it after DB save.

src/Config.ts
achal-singh: Yes, I intend to harden it after initial tests. But yeah its not a very active repo so hardening to this version should be fine.

package.json
achal-singh: Yes, I intend to harden it after initial tests. But yeah its not a very active repo so hardening to this version should be fine.

src/Config.ts
achal-singh: Yeah I initially set it this way for ease in reading the variables, can change it to camel case.

src/Config.ts
achal-singh: Oh yes this was a genuine mistake, will correct it in my next commit.

src/txDigester/ipfsPublisher.ts
achal-singh: I included the hash only to keep the log line short and the hash makes it easier to search in a DB (if someone wants to quickly look it up since its just 1 value).
I'll add in start and end cycles too.

src/txDigester/ipfsPublisher.ts
achal-singh: So I'm assuming that the upload will happen once in 10 minutes.

**BUT**, if we start the tx digester after the network has processed over, say 100 cycles, the DB will insert tx digests rapidly, triggering the **`client.uploadFile()`** function at the same rate. However, since the API from Web3.Storage isn't highly performant, frequent calls aren't ideal. To address this, I added an **`isPublisherActive`** flag and a **`REST_PERIOD_BETWEEN_UPLOADS`** (set to 5 minutes) to reduce the pressure of consecutive **`client.uploadFile()`** calls.

As an added layer of resilience, to handle the case defined above. I also thought of having a local map of all the tx digests that are to be uploaded to IPFS (something like a local cache) so that in case the **`client.uploadFile()`** errors out, which it does a lot in my recent tests, it'll be easier to ensure that we don't miss out on uploading any digests.

> Does Web3.Storage allow uploading duplicate data:

NO, it doesn't, it simply returns the CID of the data which is already published, so that's not an issue.

> Having a separate process to handle upload files to IPFS:

We can implement it, but in normal cases, there's a 10-minute window for uploads, so a separate process isn't necessary. However, if we notice CPU and memory spikes, we can create a new worker.
During stress conditions, when files need to be uploaded consecutively, we can use a **`new Map()`** as a cache or queue for tx digests.

> Do we have any retry mechanism for this?

Programmatically I haven't provisioned a retry mechanism yet. But during tests I encountered 3 different kinds of errors from Web3.Storage service's end, one of those errors is about socket connection breakage, that error automatically executes 3 retry attempts for file upload. But the other 2 types of errors do not do that. Again, I feel this can be handled using a simple Map approach as discussed above.

src/txDigester/ipfsPublisher.ts
achal-singh: Yeah we can mask it for sure, although it can't be exploited, I consoled it initially just for testing purpose.

src/txDigester/txDigestFunctions.ts
achal-singh: > What happens if we upload same cycle range to web3 storage? Will we write new data or replace existing data

Already answered this one in the big response above, in short, the same cycle range will NOT be uploaded again for the exact same data.

> Also, I am now wondering how we will be using this storage.. who is going to access, what parameters will they use to access the data.. is there mapping between txDigest record & uploaded blob id?

From what I was told by Jai, this data is being published on IPFS for record keeping purpose only, no one or no service would ping this data.

PR #87, opened 2024-09-24
Add a seperate logger for worker processes

tanuj-shardeum:


archiver-log.json
```diff
@@ -1,7 +1,7 @@
 {
     "saveConsoleOutput": true,
     "dir": "archiver-logs",
-    "files": { "main": "", "fatal": "", "net": "" },
+    "files": { "main": "", "fatal": "", "net": "", "workerProcess": "" },
```
jairajdev, line 4: Can we give the name as workerHandler instead? I think since it's more about the debug logs by the main process handling the worker processes.

PR #92, opened 2024-10-17
SHARD-51: increase GENESIS_ACCOUNTS_CYCLE_RANGE to allow debug delays for the f…

ahmxdiqbal:
…irst node

### DIFF
--- a/src/Data/Data.ts
+++ b/src/Data/Data.ts
@@ -65,7 +65,7 @@ const {
 
 const GENESIS_ACCOUNTS_CYCLE_RANGE = {
   startCycle: 0,
-  endCycle: 5,
+  endCycle: 7,
 }
 
 export enum DataRequestTypes {


urnotsam (approved):

src/Data/Data.ts
```diff
@@ -65,7 +65,7 @@ const {
 
 const GENESIS_ACCOUNTS_CYCLE_RANGE = {
   startCycle: 0,
-  endCycle: 5,
+  endCycle: 7,
```
PudgyPug, line 68: is this always going to work with 7 or is this dependent on the nodeDelay we set in our configs? If so shouldn't there be a similar nodeDelay config in the archiver that results in the correct endCycle here

ahmxdiqbal: Good catch. this was only needed for the first node, and I think its cleaner to just skip the debug delays for the first node than change GENESIS_ACCOUNTS_CYCLE_RANGE in archiver based upon it. Closing this PR

PR #93, opened 2024-10-23
SHARD-946 - Fix: By default use main program to handle global txs

jintukumardas:
- Added 2 new flags :
  disableOffloadReceipt: boolean // To disable offloading of receipts globally
  disableOffloadReceiptForGlobalModification: boolean // To disable offloading of receipts for global modifications receipts
  
- By default main program is used to handle global txs instead of workers

### DIFF
--- a/src/primary-process/index.ts
+++ b/src/primary-process/index.ts
@@ -233,6 +233,26 @@ export const offloadReceipt = async (
   receivedReceiptCount++ // Increment the counter for each receipt received
   receiptLoadTraker++ // Increment the receipt load tracker
   let verificationResult: ReceiptVerificationResult
+
+  // Check if offloading is disabled globally or for global modifications
+  if (
+    config.disableOffloadReceipt ||
+    (config.disableOffloadReceiptForGlobalModification && receipt.globalModification)
+  ) {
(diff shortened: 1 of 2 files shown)


aniketdivekar (approved):

src/primary-process/index.ts
```diff
@@ -233,6 +233,26 @@ export const offloadReceipt = async (
...
   let verificationResult: ReceiptVerificationResult
+
+  // Check if offloading is disabled globally or for global modifications
+  if (
+    config.disableOffloadReceipt ||
+    (config.disableOffloadReceiptForGlobalModification && receipt.globalModification)
+  ) {
+    mainProcessReceiptTracker++
+    if (config.workerProcessesDebugLog) console.log('Verifying on the main program', txId, timestamp)
+    verificationResult = await verifyArchiverReceipt(receipt, requiredSignatures)
```
arhamj, line 244: With these flags we still want the verifyArchiverReceipt to happen but on the main thread. This will avoid the verification all together.

arhamj: Ignore this

arhamj (approved):

PR #95, opened 2024-10-25
SHARD-1139: add better security for limit/offset query params

PudgyPug:


src/dbstore/accounts.ts
```diff
@@ -114,6 +118,10 @@ export async function queryLatestAccounts(count: number): Promise<AccountsCopy[]
 export async function queryAccounts(skip = 0, limit = 10000): Promise<AccountsCopy[]> {
   let dbAccounts: DbAccountCopy[]
   const accounts: AccountsCopy[] = []
+  if (!Number.isInteger(skip) || !Number.isInteger(limit)) {
+    Logger.mainLogger.error('Invalid skip or limit value')
```
aniketdivekar: suggestion: prefix the log w/ method name. Reason for this is, we have exact same log in `queryAccountsBetweenCycles` function also. prefix will help us differentiate.
We can/should follow this for all logs in current PR.

PudgyPug: Good point. I prefixed the logs

urnotsam (approved):

aniketdivekar (approved):

PR #96, opened 2024-10-28
SHARD-990: verifyGlobalTxReceipt unit tests

S0naliThakur:
Linear: https://linear.app/shm/issue/SHARD-990
Summary: unit tests for verifyGlobalTxReceipt method

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "homepage": "https://gitlab.com/shardus/archive/archive-server#readme",
   "devDependencies": {
-    "@types/jest": "^29.5.14",
+    "@types/jest": "29.5.14",
     "@types/node": "18.16.1",
     "@types/node-cron": "3.0.7",
     "@types/node-fetch": "^2.5.4",
@@ -69,11 +69,11 @@
     "eslint-plugin-standard": "^4.0.1",
     "eslint-plugin-xss": "^0.1.12",
     "gts": "^3.1.0",
-    "jest": "^29.7.0",
+    "jest": "29.7.0",
     "np": "8.0.4",
     "npm-run-all": "^4.1.5",
     "shx": "^0.3.4",
-    "ts-jest": "^29.2.5",
+    "ts-jest": "29.2.5",
     "typescript": "^4.4.4",
     "typescript-json-schema": "^0.51.0"
   },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -50,6 +49,7 @@
   },
   "homepage": "https://gitlab.com/shardus/archive/archive-server#readme",
   "devDependencies": {
+    "@types/jest": "^29.5.14",
```
aniketdivekar: let's please use fixed versions

src/dbstore/accounts.ts
```diff
@@ -10,7 +10,7 @@ export type AccountsCopy = {
   data: any // eslint-disable-line @typescript-eslint/no-explicit-any
   timestamp: number
   hash: string
-  cycleNumber: number
+  cycleNumber?: number
```
aniketdivekar, line 13: does this have to be optional?

S0naliThakur: yes. not all AccountsCopy objects have cycleNumber

package.json
S0naliThakur: done

aniketdivekar (approved):

tsconfig.json
```diff
@@ -2,12 +2,13 @@
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
     "target": "ES2021",
-    "rootDir": "src",
```
shyamvyas5, line 5: As we are changing the build directory, it's breaking the deployment flow as it creates `server.js` in different directory than we expect/used to be which is inside `build` directory, so we can run `node build/server.js`

shyamvyas5:
And not just that we've to change command in `package.json` as `npm run start` command expects `server.js` to be inside `build` directory

shyamvyas5:
Please keep the old flow and inside the same directory where it used to put the build before. Thanks.

tsconfig.json
shyamvyas5: We've to change command in `package.json` as well [here](https://github.com/shardeum/archive-server/blob/23e6cbb6a32d32eebca5ee1a27f5b91b34160d98/package.json#L18) `npm run start` command expects `server.js` to be inside `build` directory

PR #102, opened 2024-11-20
SHARD-1063 : Potential DoS of archiver-server during network restoration via get_account_data_archiver call

abdulazeem-tk4vr:
linear :  https://linear.app/shm/issue/SHARD-1063/[bug-bounty]-35709-potential-dos-of-archiver-server-during-network

For this task, we have added a limit on the max records that can be requested by the validator using the aforementioned API endpoint. This helps us avoid DoS vulnerabilities that may arise during the restoration of the network. If a malicious node does request for records in high magnitude, their request is rejected.

### DIFF
--- a/src/API.ts
+++ b/src/API.ts
@@ -933,6 +933,13 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage, 
       reply.send({ success: false, error: result.error })
       return
     }
+    if (payload.maxRecords > config.maxRecordsPerRequest) {
+      reply.send({
+        success: false,
+        error: 'AccountBucket Size has exceeded the max records allowed per request',
+      })
+      return
+    }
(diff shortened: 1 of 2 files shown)


src/API.ts
```diff
@@ -933,6 +933,13 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
       reply.send({ success: false, error: result.error })
       return
     }
+    if (payload.maxRecords > config.maxRecordsPerRequest) {
+      reply.send({
+        success: false,
+        error: 'AccountBucket Size has exceeded the max records allowed per request',
```
aniketdivekar: can we also add `max number` that we allow in error message?

aniketdivekar (approved):

jintukumardas (approved):

PR #103, opened 2024-11-21
SHARD-44: GlobalTx fixes

S0naliThakur:
Linear: https://linear.app/shm/issue/SHARD-44/globaltx-validation

Summary: Implement global receipt validation and enhance receipt verification logic

1. Put archiver robust query behind `useRobustQueryForReceipt` flag
2. Integrated AJV schema validation for global receipts across multiple modules.
3. Improved error handling and logging for receipt validation processes.
4. Updated AJV schema enumeration to include GlobalTxReceipt.
5. Modified `storeReceipt` method to use `verifyArchiverReceipt` directly.
6. Added functions for verifying global receipts, including `verifyGlobalTxreceiptOffline` and `verifyReceiptMajority` for better code readability.
7. Refactored existing receipt verification methods to support both global and non-global receipts.

### DIFF
--- a/src/Data/Collector.ts
+++ b/src/Data/Collector.ts
@@ -67,7 +67,7 @@ const verifyReceiptMajority = async (
   receipt: Receipt.Receipt | Receipt.ArchiverReceipt,
   executionGroupNodes: ConsensusNodeInfo[],
   minConfirmations: number = config.RECEIPT_CONFIRMATIONS
-): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt }> => {
+): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt | Receipt.Receipt }> => {
   /**
    * Note:
    * Currently, only the non-global receipt flow is implemented in `verifyReceiptMajority`,
@@ -101,7 +101,7 @@ const verifyReceiptOffline = async (
   receipt: Receipt.Receipt | Receipt.ArchiverReceipt,
   executionGroupNodes: ConsensusNodeInfo[],
   minConfirmations: number
-): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt }> => {
+): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt | Receipt.Receipt }> => {
   return verifyNonGlobalTxReceiptOffline(receipt, executionGroupNodes, minConfirmations)
 }
 
@@ -117,7 +117,7 @@ const verifyReceiptWithValidators = async (
   receipt: Receipt.Receipt | Receipt.ArchiverReceipt,
   executionGroupNodes: ConsensusNodeInfo[],
   minConfirmations: number = config.RECEIPT_CONFIRMATIONS
-): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt }> => {
+): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt | Receipt.Receipt }> => {
   return verifyNonGlobalTxReceiptWithValidators(receipt, executionGroupNodes, minConfirmations)
 }
 
@@ -132,7 +132,7 @@ const verifyNonGlobalTxReceiptWithValidators = async (
   receipt: Receipt.Receipt | Receipt.ArchiverReceipt,
   executionGroupNodes: ConsensusNodeInfo[],
   minConfirmations: number = config.RECEIPT_CONFIRMATIONS
-): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt }> => {
+): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt | Receipt.Receipt }> => {
   const result = { success: false }
   // Created signedData with full_receipt = false outside of queryReceipt to avoid signing the same data multiple times
   let signedData = Crypto.sign({
@@ -253,7 +253,7 @@ const verifyNonGlobalTxReceiptWithValidators = async (
         )
         continue
       }
-      const fullReceipt = fullReceiptResult.receipt as Receipt.ArchiverReceipt
+      const fullReceipt = fullReceiptResult.receipt as Receipt.ArchiverReceipt | Receipt.Receipt
       if (
         isReceiptEqual(fullReceipt.signedReceipt, robustQueryReceipt) &&
         validateReceiptType(fullReceipt)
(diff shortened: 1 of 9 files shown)


src/Data/Collector.ts
```diff
@@ -270,6 +275,87 @@ const isReceiptRobust = async (
...
+    // Ensure tx exists and has required fields
+    if (!globalReceipt.tx || !globalReceipt.tx.address) {
+      return { success: false }
+    }
+
+    for (const sign of globalReceipt.signs) {
+      // Skip invalid signatures
+      if (!sign || !sign.owner) continue
+
+      const node = executionGroupNodes.find(n => n.publicKey === sign.owner)
```
aniketdivekar: should this comparison be case insensitive?

src/Data/Collector.ts
```diff
@@ -270,6 +275,87 @@ const isReceiptRobust = async (
...
+  const validSigners = new Set<string>()
+
+  if (!normalReceipt.signaturePack || !Array.isArray(normalReceipt.signaturePack)) {
+    return { success: false }
+  }
+
+  for (const signature of normalReceipt.signaturePack) {
+    if (!signature || !signature.owner) continue
+
+    const node = executionGroupNodes.find(n => n.publicKey === signature.owner)
```
aniketdivekar: same as above

src/Data/Collector.ts
```diff
@@ -270,6 +275,87 @@ const isReceiptRobust = async (
...
+      const appliedVoteHash = {
+        txid: receipt.tx.txId,
+        voteHash
+      }
+
+      if (Crypto.verify({ ...appliedVoteHash, sign: signature })) {
+        validSigners.add(signature.owner)
+      }
+    } catch (error) {
+      console.error('Error verifying signature:', error)
```
aniketdivekar, line 450: error message in global tx [validation](https://github.com/shardeum/archive-server/pull/103/files#diff-186c711dbe48d8ffd4c6488e2d29090dae64abd946a5cf185a549af2c556a72eR313) is also the same. we can use differentiator text like `globalTx` & `nonGlobalTx`

src/Data/Collector.ts
```diff
@@ -270,6 +275,87 @@ const isReceiptRobust = async (
...
 
+// Offline receipt verification
+const verifyReceiptOffline = async (
+  receipt: Receipt.ArchiverReceipt, 
+  executionGroupNodes: ConsensusNodeInfo[],
+  minConfirmations: number
+): Promise<{ success: boolean; newReceipt?: Receipt.ArchiverReceipt }> => {
+  const { signedReceipt, globalModification } = receipt
+
+  if (globalModification) {
```
aniketdivekar: it will be more readable if we refactor this method to have two more function invoked based on this flag; `verifyGlobalTxReceiptOffline` and `verifyNonGlobalTxReceiptOffline`

abdulazeem-tk4vr: we could probably call it internally instead of calling it from the main function

abdulazeem-tk4vr: we could probably call it internally for different methods instead of calling it twice from the main function

src/Data/Collector.ts
```diff
@@ -270,6 +275,87 @@ const isReceiptRobust = async (
   return { success: true }
 }
 
+// Offline receipt verification
+const verifyReceiptOffline = async (
+  receipt: Receipt.ArchiverReceipt,
```
abdulazeem-tk4vr: small nitpick, formatting is needed

src/Data/Collector.ts
S0naliThakur: changed in updated code

src/Data/Collector.ts

devendra-shardeum (dismissed):

aniketdivekar (dismissed):

jintukumardas:
@S0naliThakur  Since we have disabled the workers. We can also comment out the code where we are initializing and setting up the worker process. Please refer to these lines - 
1. https://github.com/shardeum/archive-server/blob/780ed888bb31bba232e5c79ac02ae8db50bf9a6d/src/server.ts#L96
2. https://github.com/shardeum/archive-server/blob/780ed888bb31bba232e5c79ac02ae8db50bf9a6d/src/server.ts#L500

CC: @aniketdivekar

jintukumardas (changes requested):
Please comment out the initWorkerProcess and setupWorkerProcesses methods.

aniketdivekar (dismissed):

jintukumardas (dismissed):

aniketdivekar (dismissed):

src/Data/Collector.ts
```diff
@@ -272,6 +329,134 @@ const isReceiptRobust = async (
...
+      nestedCountersInstance.countEvent(
+        'receipt',
+        `Invalid_receipt_globalModification_signs_count_less_than_${config.requiredMajorityVotesPercentage}%`
+      )
+    return result
+  }
+
+  const nodeMap = new Map<string, P2PTypes.NodeListTypes.Node>()
+  // Fill the map with nodes keyed by their public keys
+  cycleShardData.nodes.forEach((node) => {
```
afostr, line 370: I think it would be ok to just put the nodeMap on the cycleShardData the first time we compute it (lazy init) 
that does not need to change to merge this.  I suspect we will do a more general perf pass later as time permits.

src/Data/Collector.ts
```diff
@@ -272,6 +329,134 @@ const isReceiptRobust = async (
...
+      Logger.mainLogger.error(
+        `The node with public key ${nodePubKey} of the receipt ${txId} with ${timestamp} is not in the active nodesList of cycle ${cycle}`
+      )
+      if (nestedCountersInstance)
+        nestedCountersInstance.countEvent('receipt', 'globalModification_sign_owner_not_in_active_nodesList')
+      continue
+    }
+    // Check if the node is in the execution group
+    if (!cycleShardData.parititionShardDataMap.get(homePartition).coveredBy[node.id]) {
+      Logger.mainLogger.error(
```
afostr, line 391: I think in the future we should generally be putting log lines behind if checks like in the validator.  This is because even if you don't log it you still pay the string format cost.
(this is not a blocker though)

afostr: I do have some logger pattern upgrades to share later anyhow.

src/Data/Collector.ts
```diff
@@ -272,6 +329,134 @@ const isReceiptRobust = async (
...
+  const validSigners = new Set<string>()
+
+  if (!normalReceipt.signaturePack || !Array.isArray(normalReceipt.signaturePack)) {
+    return { success: false }
+  }
+
+  for (const signature of normalReceipt.signaturePack) {
+    if (!signature || !signature.owner) continue
+
+    const node = executionGroupNodes.find((n) => n.publicKey.toLowerCase() === signature.owner.toLowerCase())
```
afostr, line 436: when we later do a perf pass a map would avoid O(n^2)

afostr (dismissed):
I made a few small comments but they are not show stoppers to get into dev.  I think the testing and roll out plans will be critical for this

aniketdivekar (dismissed):

aniketdivekar (approved):

devendra-shardeum (approved):
looks good now

PR #109, opened 2024-12-08
added global error handling for archive-server and handling for unhandledRejections in worker-processes

devendra-shardeum:


mssabr01 (approved):
Approved. We must strive to keep track of the errors caught this way and be vigilant about catching those errors locally. That process is out of scope for these changes though

PR #115, opened 2024-12-26
SEC-531 : Adding robustQuery to fetch secure accounts from active validators

abdulazeem-tk4vr:
Linear : https://linear.app/shm/issue/SEC-531/adding-robustquery-to-fetch-secure-accounts-from-active-validators

### DIFF
--- a/src/Data/ExtractShardKey.ts
+++ b/src/Data/ExtractShardKey.ts
@@ -1,6 +1,44 @@
 import { config } from '../Config'
 import { AccountType } from '../shardeum/calculateAccountHash'
 import { InternalTXType } from '../shardeum/verifyGlobalTxReceipt'
+import * as NodeList from '../NodeList'
+import { getJson } from '../P2P'
+import * as Utils from '../Utils'
+
+export async function crackKeyFromSecureAccount(tx: any): Promise<string> {
+  // Define the query function for robustQuery
+  const queryFn = async (node: NodeList.ConsensusNodeInfo): Promise<any[]> => {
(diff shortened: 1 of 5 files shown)


src/Data/ExtractShardKey.ts
```diff
@@ -1,6 +1,44 @@
...
+  )
+
+  if (!matchingAccount) {
+    throw new Error(`Secure account ${tx.accountName} not found`)
+  }
+
+  const secureAccountConfig = matchingAccount.secureAccountConfig
+
+  // Return the SourceFundsAddress formatted as a Shardus address
+  return toShardusAddress(secureAccountConfig.SourceFundsAddress, AccountType.Account)
```
aniketdivekar, line 40: should the account type be `Account` or `SecureAccount`?

abdulazeem-tk4vr:
```
export function crack(tx: InternalTx): CrackedData {
  if (!secureAccountDataMap.has(tx.accountName)) {
    throw new Error(`Secure account ${tx.accountName} not found`);
  }
  return {
    sourceKeys: [
      toShardusAddress(secureAccountDataMap.get(tx.accountName).SourceFundsAddress, AccountType.Account),
      toShardusAddress(secureAccountDataMap.get(tx.accountName).SecureAccountAddress, AccountType.SecureAccount),
    ],
    targetKeys: [
      toShardusAddress(secureAccountDataMap.get(tx.accountName).RecipientFundsAddress, AccountType.Account)
    ]
  }
}
```

The first key is of type account which is relevant to us

abdulazeem-tk4vr: It should be `Account`

PR #116, opened 2024-12-26
SEC-530 : Adding getTxSenderAddress to the archiver

abdulazeem-tk4vr:
Linear: https://linear.app/shm/issue/SEC-530/adding-gettxsenderaddress-to-the-archiver

### DIFF
--- a/src/Data/ExtractShardKey.ts
+++ b/src/Data/ExtractShardKey.ts
@@ -4,6 +4,15 @@ import { InternalTXType } from '../shardeum/verifyGlobalTxReceipt'
 import * as NodeList from '../NodeList'
 import { getJson } from '../P2P'
 import * as Utils from '../Utils'
+import { Address } from '@ethereumjs/util'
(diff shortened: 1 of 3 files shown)


S0naliThakur (approved):

src/Data/ExtractShardKey.ts
```diff
@@ -148,6 +157,56 @@ async function extractKeyFromInternalTx(tx: any): Promise<string> {
...
+        return cached
+      }
+    }
+
+    const rawTx = '0x' + toHexString(tx.serialize())
+    const { address, isValid, gasValid } = getSenderAddress(rawTx)
+    if (config.VERBOSE) console.log('Sender address retrieved from signed txn', address)
+    const res = { address: Address.fromString(address), isValid, gasValid }
+    if (txid != null) {
+      simpleTTL++
```
aniketdivekar, line 192: should this be only incremented if txid is not in txSenderCache?
Also, if we are clearing the cache, shouldn't we set simpleTTL to 0?

PR #118, opened 2024-12-26
SEC-526

abdulazeem-tk4vr:
Linear : https://linear.app/shm/issue/SEC-526/archiver-is-not-checking-list-of-accounts-belonging-to-actual-accounts


src/Data/Collector.ts
```diff
@@ -298,8 +299,21 @@ export const verifyReceiptData = async (
...
+
+  if (config.enableKeyExtraction) {
+    if (config.VERBOSE) console.log('The Incoming executionShardKey key is', executionShardKey)
+    const extractedKey: string = await extractKeyFromTx(receipt.tx)
+    if (extractedKey != executionShardKey) {
+      if (config.VERBOSE) {
+        console.log('Execution Shard Key in receipt does not match the calculated key from transaction data.')
+      }
+      if (nestedCountersInstance) nestedCountersInstance.countEvent('receipt', 'executionShardKey mismatch')
+      executionShardKey = extractedKey
```
aniketdivekar, line 517: I wonder if fallback should be to the key which we received in receipt. Just in case our extraction logic has some flaws. We can discuss with design team.

src/Data/ExtractShardKey.ts
```diff
@@ -0,0 +1,259 @@
...
+export async function crackKeyFromSecureAccount(tx: any): Promise<string> {
+  // Define the query function for robustQuery
+  const queryFn = async (node: NodeList.ConsensusNodeInfo): Promise<any[]> => {
+    const response = (await getJson(`http://${node.ip}:${node.port}/secure_accounts`)) as any
+
+    if (response.accounts) {
+      return response.accounts
+    }
+
+    throw new Error(`Secure account data not found on node ${node.ip}`)
```
aniketdivekar: can we add some logs here?if there is any error in response from /secure_accounts api, we lose the error message.

src/test/shardKey.ts
```diff
@@ -0,0 +1,32 @@
+import * as fs from 'fs'
```
aniketdivekar, line 1: can this be written as unit test? what's the purpose of the file? 
Also, `filePath` is hardcoded

src/Data/Collector.ts
abdulazeem-tk4vr: Well, I was thinking about the possibility that the incoming execution shard key from the receipt might contain malicious/falsified information. So, we should instead rely on the shard key we have calculated now.

src/Data/ExtractShardKey.ts
abdulazeem-tk4vr: this will not be used anymore. We are going to use a local config file to fetch the secure accounts

src/test/shardKey.ts
abdulazeem-tk4vr: There are multiple different unit tests to cover because of different transaction types and the receipt logs are quite large. I thought this was the most efficient way to test out the function by fetching the receipt logs from the archiver. Testing is still incomplete, QA help is needed. 

FilePath was just an example. It can be modified to anything you like. I shall keep it generic and add a disclaimer that this needs to be modified.

PR #123, opened 2025-01-03
SHARD-1166: Archiver whitelisting

jintukumardas:
Use the signed list of allowed archivers.

### DIFF
--- a/allowed-archivers.json
+++ b/allowed-archivers.json
@@ -14,12 +14,11 @@
     "allowedAccounts": {
         "0x002D3a2BfE09E3E29b6d38d58CaaD16EEe4C9BC5": 3
     },
-    "counter": 1,
     "minSigRequired": 1,
     "signatures": [
         {
             "owner": "0x002D3a2BfE09E3E29b6d38d58CaaD16EEe4C9BC5",
-            "sig": "0xc4f25a2156c9151c0515d1e5636b94de5ab2e21cc924207a2d2b5d679d8b3dec0f3b7aab78a0dfa444ee4794cf3a61aa6b9a70f8a51ebac46b9560d50cb019b51b"
+            "sig": "0x53535921e57d0796b0dbf1451ff6a9ff535bbf2f51cdcc1a5ddd618aa62ca0f36a98c5e2fcecaa7217626735fc3bb5f47acc5e7e60284bb20fb157b50390d45f1b"
         }
     ]
 }
--- a/allowed-archivers.json
+++ b/allowed-archivers.json
@@ -11,11 +11,15 @@
             "publicKey": "e8a5c26b9e2c3c31eb7c7d73eaed9484374c16d983ce95f3ab18a62521964a94"
         }
     ],
+    "allowedAccounts": [
+        "0x002D3a2BfE09E3E29b6d38d58CaaD16EEe4C9BC5"
+    ],
+    "counter": 1,
+    "minSigRequired": 1,
     "signatures": [
         {
             "owner": "0x002D3a2BfE09E3E29b6d38d58CaaD16EEe4C9BC5",
-            "sig": "0xf330bf98a7154ddbb5472d3f1cb3153f0c55f9ad6625d456ec5b99ce9e0701770910db7fab0a697388b587d9df1ebc6bc035793ac282b02c097482c37568213f1b"
+            "sig": "0x95b32d241e1885b9db2f2a5850c99a10a6a1df7f8142b0914715a1a5bdfd9bed09753f32948cb9fee69b20cfefee89d3f90be183de721d1b5d88157a6affda821b"
         }
-    ],
-    "counter": 1
+    ]
 }
--- a/allowed-archivers.json
+++ b/allowed-archivers.json
@@ -1,0 +1,21 @@
+{
+    "allowedArchivers": [
+        {
+            "ip": "127.0.0.1",
+            "port": 4000,
+            "publicKey": "758b1c119412298802cd28dbfa394cdfeecc4074492d60844cc192d632d84de3"
+        },
+        {
+            "ip": "127.0.0.1",
+            "port": 4001,
+            "publicKey": "e8a5c26b9e2c3c31eb7c7d73eaed9484374c16d983ce95f3ab18a62521964a94"
+        }
+    ],
+    "signatures": [
+        {
+            "owner": "0x002D3a2BfE09E3E29b6d38d58CaaD16EEe4C9BC5",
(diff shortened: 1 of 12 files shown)


allowed-archivers.json
```diff
@@ -0,0 +1,21 @@
...
+        {
+            "ip": "127.0.0.1",
+            "port": 4001,
+            "publicKey": "e8a5c26b9e2c3c31eb7c7d73eaed9484374c16d983ce95f3ab18a62521964a94"
+        }
+    ],
+    "signatures": [
+        {
+            "owner": "0x002D3a2BfE09E3E29b6d38d58CaaD16EEe4C9BC5",
+            "sig": "0xf330bf98a7154ddbb5472d3f1cb3153f0c55f9ad6625d456ec5b99ce9e0701770910db7fab0a697388b587d9df1ebc6bc035793ac282b02c097482c37568213f1b"
```
jintukumardas: this is not the proper signature.

src/API.ts
```diff
@@ -294,6 +297,19 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
       request.raw.socket.destroy()
     }
   })
+  server.get('/allowed-archivers', async (request, reply) => {
+    try {
+      const allowedArchiversPath = path.resolve(__dirname, '../allowed-archivers.json')
```
jintukumardas: May add a file watcher but in local tests I found that it sends the latest file

src/API.ts
```diff
@@ -1296,19 +1369,94 @@ export const validateRequestData = (
...
-        return { success: false, error: 'Data request sender is not an archiver' }
-      }
-    }
+
     if (!Crypto.verify(data)) {
       Logger.mainLogger.error('Invalid signature', data)
       return { success: false, error: 'Invalid signature' }
     }
+
+    if (!skipArchiverCheck && config.limitToArchiversOnly) {
```
jintukumardas, line 1319: should we remove the limitToArchiversOnly flag? Not sure about the intention behind this flag.

src/API.ts
```diff
@@ -281,6 +287,73 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+      // Load and parse initial config
+      const loadConfig = () => {
+        const data = fs.readFileSync(allowedArchiversPath, 'utf8')
+        return StringUtils.safeJsonParse(data)
+      }
+
+      let allowedArchivers = loadConfig()
+
+      // Watch for config file changes
+      fs.watchFile(allowedArchiversPath, (curr, prev) => {
```
aniketdivekar: this is being done in every API call? won't that create multiple event watchers?
I think this should not be part of API. this should be a periodic invocation on background and validating the archivers in the file. We can maintain that file/list of archiver in memory. In-memory list will always be valid. it will be generated on startup and updated on file change if signatures are valid. this API should simply return the list in memory(with fallback of loading again)

jintukumardas: Thanks for reviewing it. I'll move the business logic to a separate file/method.

src/API.ts
```diff
@@ -1296,19 +1369,94 @@ export const validateRequestData = (
...
+
     if (!Crypto.verify(data)) {
       Logger.mainLogger.error('Invalid signature', data)
       return { success: false, error: 'Invalid signature' }
     }
+
+    if (!skipArchiverCheck && config.limitToArchiversOnly) {
+      try {
+        // Load and validate allowed archivers config
+        const allowedArchiversPath = path.resolve(__dirname, '../allowed-archivers.json')
```
jintukumardas: WIP -  Move the verification logic to a separate file/method.

src/API.ts
jintukumardas: Done

src/API.ts
jintukumardas: Done

allowed-archivers.json
```diff
@@ -0,0 +1,25 @@
...
+            "port": 4000,
+            "publicKey": "758b1c119412298802cd28dbfa394cdfeecc4074492d60844cc192d632d84de3"
+        },
+        {
+            "ip": "127.0.0.1",
+            "port": 4001,
+            "publicKey": "e8a5c26b9e2c3c31eb7c7d73eaed9484374c16d983ce95f3ab18a62521964a94"
+        }
+    ],
+    "allowedAccounts": {
```
aniketdivekar: lets remove this and also minSigRequired

aniketdivekar: @jintukumardas this is still pending

src/shardeum/allowedArchiversManager.ts
```diff
@@ -0,0 +1,164 @@
...
+    signatures: Sign[]
+}
+
+class AllowedArchiversManager {
+    private currentConfig: AllowedArchiversConfig | null = null
+    private configPath: string = ''
+    private isInitialized: boolean = false
+    private useGlobalAccount: boolean = false
+    private globalAccountAllowedSigners: { [key: string]: number } = {}
+    private globalAccountMinSigRequired: number = 0
```
aniketdivekar, line 23: @jintukumardas shouldn't this be at least 1?

src/shardeum/allowedArchiversManager.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public setGlobalAccountConfig(allowedSigners?: { [key: string]: DevSecurityLevel }, minSigRequired?: number): void {
+        if (allowedSigners) {
+            this.globalAccountAllowedSigners = allowedSigners
+        }
+        if (minSigRequired >= 1) {
+            this.globalAccountMinSigRequired = minSigRequired
+        }
+        if (!allowedSigners && !minSigRequired) {
```
aniketdivekar, line 73: this should be OR condition right? if any of allowedSigners & minSigRequired are missing then we return?

allowed-archivers.json
jintukumardas: This is necessary for the initial config load, as the Global account data will initially be empty. This data will serve as a reference, but once we receive the Global account data, it will no longer be needed/referenced.

src/shardeum/allowedArchiversManager.ts
jintukumardas: This should be an AND condition because we can update minSigRequired without updating the multi-sig accounts. But one of them needs to present.

src/shardeum/allowedArchiversManager.ts
jintukumardas: since this value will be updated from the global account; 0 or 1 wouldn't matter.

aniketdivekar (dismissed):

aniketdivekar (approved):

PR #126, opened 2025-01-10
SHARD-1708: Add network account listOfChanges update script

S0naliThakur:
Linear: https://linear.app/shm/issue/SHARD-1708
Summary: Add network account listOfChanges update script

### DIFF
--- a/scripts/update_network_account_listOfChanges.ts
+++ b/scripts/update_network_account_listOfChanges.ts
@@ -84,12 +84,22 @@ const askConfigQuestions = async (): Promise<ConfigSchema> => {
         const expectedType = typeof defaultConfig[section][key]
         const parsedValue = parseInput(value, expectedType, defaultConfig[section][key])
 
-        if (parsedValue !== defaultConfig[section][key] || value === '') {
+        // If empty input, use default value
+        if (value === '') {
+          userConfig[section][key] = defaultConfig[section][key]
+          isValid = true
+        }
+        // If valid parsed value
+        else if (parsedValue !== undefined) {
           userConfig[section][key] = parsedValue
           isValid = true
-        } else {
+        }
+        // Invalid input
+        else {
           attempts++
-          console.log(`Invalid input. Please enter a valid ${expectedType} for ${section}.${key}. (${attempts}/${maxAttempts} attempts)`)
+          console.log(
+            `Invalid input. Please enter a valid ${expectedType} for ${section}.${key}. (${attempts}/${maxAttempts} attempts)`
+          )
         }
       }
 
@@ -112,7 +122,11 @@ const getCycleNumber = async (): Promise<number> => {
   let cycleNumberInput: number
   do {
     cycleNumberInput = parseInt(
-      await askQuestion('Enter cycle number (must be less than the latest cycle number, latest cycle number is: ' + latestCycleNumber + '): '),
+      await askQuestion(
+        'Enter cycle number (must be less than the latest cycle number, latest cycle number is: ' +
+          latestCycleNumber +
+          '): '
+      ),
       10
     )
     if (isNaN(cycleNumberInput) || cycleNumberInput >= latestCycleNumber) {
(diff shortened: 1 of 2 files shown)


scripts/update_network_account_listOfChanges.ts
```diff
@@ -0,0 +1,190 @@
...
+  input: process.stdin as NodeJS.ReadableStream,
+  output: process.stdout as NodeJS.WritableStream,
+})
+
+const askQuestion = (query: string): Promise<string> => {
+  return new Promise((resolve) => rl.question(query, resolve))
+}
+
+// Utility function to validate and parse input based on expected type
+const parseInput = (value: string, expectedType: string, defaultValue: any): any => {
```
mhanson-github, line 55: I am not sure how I feel about assigning a defaultValue.  If we're using this code to overwrite the network account, I somewhat feel that there should only be intentional and explicit changes.

scripts/update_network_account_listOfChanges.ts
```diff
@@ -0,0 +1,190 @@
...
+    networkAccount.hash = calculatedAccountHash
+    networkAccount.data.hash = calculatedAccountHash
+
+    await AccountDB.insertAccount(networkAccount)
+    console.log('✅ Changes were applied.')
+  } else {
+    console.log('❌ Changes were not applied.')
+  }
+
+  await dbstore.closeDatabase()
```
mhanson-github: Should we have a try/finally setup here to ensure DB closure?

urnotsam (dismissed):

scripts/update_network_account_listOfChanges.ts
mhanson-github: Clarifying...  the question is should we treat the results differently if we ended up with value or defaultValue?

scripts/update_network_account_listOfChanges.ts
```diff
@@ -0,0 +1,190 @@
...
+      let attempts = 0
+      const maxAttempts = 3
+      let isValid = false
+
+      while (!isValid && attempts < maxAttempts) {
+        const value = await askQuestion(`Enter ${section}.${key} (default: ${defaultConfig[section][key]}): `)
+        const expectedType = typeof defaultConfig[section][key]
+        const parsedValue = parseInput(value, expectedType, defaultConfig[section][key])
+
+        if (parsedValue !== defaultConfig[section][key] || value === '') {
```
aniketdivekar: `parsedValue == defaultConfig[section][key]` this can be a genuine case right? I mean not an error. We can suppress this error or simply say you have not updated the value.

Also what's the purpose of `value === ''` condition?

scripts/update_network_account_listOfChanges.ts
```diff
@@ -0,0 +1,190 @@
...
+// Function to get the latest cycle from db
+const getCycleNumber = async (): Promise<number> => {
+  const latestCycle = await CycleDB.queryLatestCycleRecords(1)
+  const latestCycleRecord = latestCycle[0]
+  const latestCycleNumber = latestCycleRecord.counter
+
+  let cycleNumberInput: number
+  do {
+    cycleNumberInput = parseInt(
+      await askQuestion('Enter cycle number (must be less than the latest cycle number, latest cycle number is: ' + latestCycleNumber + '): '),
```
aniketdivekar: why this has to be lesser and greater? won't the change be applied in future cycles?

scripts/update_network_account_listOfChanges.ts
S0naliThakur: The default values are there to make it quicker and easier to input values from the terminal. Users are always asked what they want to use as input, so the default values get replaced with whatever they enter. Of course, you can also tweak the inputs directly in the script if you wish. The parameters are completely flexible and can be changed as needed.
Right now, we’re keeping it simple by starting with just baselineNodes, minNodes, and maxNodes

scripts/update_network_account_listOfChanges.ts
S0naliThakur: The value === '' condition is checking for when the user just hits enter without typing anything, indicating they want to accept the default value.

scripts/update_network_account_listOfChanges.ts
S0naliThakur: This is implemented as per the ticket specification

scripts/update_network_account_listOfChanges.ts
S0naliThakur: added

aniketdivekar (approved):

urnotsam (approved):

PR #134, opened 2025-02-05
SHARD-1385: Archiver Hardening

S0naliThakur:
Linear: https://linear.app/shm/issue/SHARD-1385

Summary: Archiver Hardening v1 + fix for https://linear.app/shm/issue/SHARD-1389

### DIFF
--- a/src/API.ts
+++ b/src/API.ts
@@ -1532,12 +1532,20 @@ export const queryFromArchivers = async (
   let bucketID: number | undefined;
   const params = queryParameters as any;
   let checkpointStatusType: CheckpointStatusType | undefined;
+  let bucketRange: { start: number; end: number } | undefined;
 
   // Extract bucket ID based on the request type and parameters
   switch (type) {
     case RequestDataType.CYCLE:
       // For cycle info, use start or end as bucket ID
-      if (params.start !== undefined) {
+      if (params.start !== undefined && params.end !== undefined) {
+        // If both start and end are provided, we'll check the entire range
+        bucketRange = {
+          start: params.start,
+          end: params.end
+        };
+        bucketID = params.start; // Default to start for single bucket check fallback
+      } else if (params.start !== undefined) {
         bucketID = params.start;
       } else if (params.end !== undefined) {
         bucketID = params.end;
@@ -1553,7 +1561,14 @@ export const queryFromArchivers = async (
     case RequestDataType.ACCOUNT:
     case RequestDataType.TRANSACTION:
       // For other data types, check if startCycle or endCycle is specified
-      if (params.startCycle !== undefined) {
+      if (params.startCycle !== undefined && params.endCycle !== undefined) {
+        // If both startCycle and endCycle are provided, we'll check the entire range
+        bucketRange = {
+          start: params.startCycle,
+          end: params.endCycle
+        };
+        bucketID = params.startCycle; // Default to startCycle for single bucket check fallback
+      } else if (params.startCycle !== undefined) {
         bucketID = params.startCycle;
       } else if (params.endCycle !== undefined) {
         bucketID = params.endCycle;
@@ -1592,35 +1607,84 @@ export const queryFromArchivers = async (
   const randomArchivers = Utils.getRandomItemFromArr(State.otherArchivers, 0, maxNumberofArchiversToRetry)
   let retry = 0
 
-  // If checkpoint is enabled and we have a bucket ID, try to find an archiver with verified data first
-  if (config.checkpoint.bucketConfig.allowCheckpointUpdates && bucketID !== undefined && randomArchivers.length > 0) {
-    for (const archiver of randomArchivers) {
-      if (!archiver) continue;
-      try {
-        // Check if this archiver has verified the bucket
-
-        const data = {
-          bucketID: bucketID.toString(),
-          sender: config.ARCHIVER_PUBLIC_KEY,
+  // This logic is not optimized, we should only check the bucket verification once and then use the verified archiver to get the data
+  // If checkpoint is enabled and we have a bucket range or bucket ID, try to find an archiver with verified data first
+  if (config.checkpoint.bucketConfig.allowCheckpointUpdates && randomArchivers.length > 0) {
+    if (bucketRange !== undefined) {
+      // Check verification for the entire range of buckets
+      for (const archiver of randomArchivers) {
+        if (!archiver) continue;
+        
+        let allBucketsVerified = true;
+        
+        // Check each bucket in the range
+        for (let currentBucket = bucketRange.start; currentBucket <= bucketRange.end; currentBucket++) {
+          try {
+            const verificationData = {
+              bucketID: currentBucket.toString(),
+              sender: config.ARCHIVER_PUBLIC_KEY,
+            }
+            const signedBucketVerificationDataToSend = Crypto.sign(verificationData)
+            const verificationResponse: { success?: boolean, isVerified?: boolean } = await P2P.postJson(
+              `http://${archiver.ip}:${archiver.port}/bucket-verification`,
+              signedBucketVerificationDataToSend,
+              timeoutInSecond
+            )
+            
+            if (!(verificationResponse?.success && verificationResponse?.isVerified)) {
+              allBucketsVerified = false;
+              break; // If any bucket is not verified, move to the next archiver
+            }
+          } catch (error) {
+            Logger.mainLogger.error(`Error checking bucket verification for bucket ${currentBucket} on archiver ${archiver.ip}:${archiver.port}:`, error);
+            allBucketsVerified = false;
+            break;
+          }
         }
-        const signedBucketVerificationDataToSend = Crypto.sign(data)
-        const verificationResponse: { success?: boolean, isVerified?: boolean } = await P2P.postJson(
-          `http://${archiver.ip}:${archiver.port}/bucket-verification`,
-          signedBucketVerificationDataToSend,
-          timeoutInSecond
-        )
-        const isVerifiedBucket = verificationResponse?.success && verificationResponse?.isVerified
-        if (isVerifiedBucket) {
-          // Try to get data from this archiver since it has verified data
-          const response = await P2P.postJson(
-            `http://${archiver.ip}:${archiver.port}${url}`,
-            signedDataToSend,
+        
+        if (allBucketsVerified) {
+          // All buckets in the range are verified, use this archiver
+          try {
+            const response = await P2P.postJson(
+              `http://${archiver.ip}:${archiver.port}${url}`,
+              signedDataToSend,
+              timeoutInSecond
+            )
+            if (response) return response;
+          } catch (error) {
+            Logger.mainLogger.error(`Error querying data from verified archiver ${archiver.ip}:${archiver.port}:`, error);
+          }
+        }
+      }
+    } else if (bucketID !== undefined) {
+      // Original single bucket verification logic
+      for (const archiver of randomArchivers) {
+        if (!archiver) continue;
+        try {
+          // Check if this archiver has verified the bucket
+          const data = {
+            bucketID: bucketID.toString(),
+            sender: config.ARCHIVER_PUBLIC_KEY,
+          }
+          const signedBucketVerificationDataToSend = Crypto.sign(data)
+          const verificationResponse: { success?: boolean, isVerified?: boolean } = await P2P.postJson(
+            `http://${archiver.ip}:${archiver.port}/bucket-verification`,
+            signedBucketVerificationDataToSend,
             timeoutInSecond
           )
-          if (response) return response;
+          const isVerifiedBucket = verificationResponse?.success && verificationResponse?.isVerified
+          if (isVerifiedBucket) {
+            // Try to get data from this archiver since it has verified data
+            const response = await P2P.postJson(
+              `http://${archiver.ip}:${archiver.port}${url}`,
+              signedDataToSend,
+              timeoutInSecond
+            )
+            if (response) return response;
+          }
+        } catch (error) {
+          Logger.mainLogger.error(`Error checking bucket verification for archiver ${archiver.ip}:${archiver.port}:`, error);
         }
-      } catch (error) {
-        Logger.mainLogger.error(`Error checking bucket verification for archiver ${archiver.ip}:${archiver.port}:`, error);
       }
     }
   }
--- a/src/API.ts
+++ b/src/API.ts
@@ -1418,13 +1418,13 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage, 
   })
 
   type BucketVerificationRequest = FastifyRequest<{
-    Body: { bucketID: string }
+    Body: { bucketID: string; endBucketID?: string }
   }>
 
   server.post('/bucket-verification', async (request: BucketVerificationRequest & Request, reply) => {
     try {
       const requestData = request.body
-      const result = validateRequestData(requestData, { bucketID: 's', sender: 's', sign: 'o' })
+      const result = validateRequestData(requestData, { bucketID: 's', endBucketID: '?s', sender: 's', sign: 'o' })
       if (!result.success) {
         reply.code(400).send({
           success: false,
@@ -1434,6 +1434,18 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage, 
       }
 
       const bucketID = parseInt(requestData.bucketID, 10)
+      let endBucketID: number | undefined = undefined
+
+      if (requestData.endBucketID) {
+        endBucketID = parseInt(requestData.endBucketID, 10)
+        if (isNaN(endBucketID) || endBucketID < 0) {
+          reply.code(400).send({
+            success: false,
+            error: 'Invalid endBucketID. Must be a non-negative integer.'
+          })
+          return
+        }
+      }
 
       if (isNaN(bucketID) || bucketID < 0) {
         reply.code(400).send({
@@ -1443,7 +1455,7 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage, 
         return
       }
 
-      const isVerified = await isBucketVerified(bucketID)
+      const isVerified = await isBucketVerified(bucketID, endBucketID)
 
       const response = Crypto.sign({
         success: true,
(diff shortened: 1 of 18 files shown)


src/API.ts
```diff
@@ -216,6 +223,28 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
     profilerInstance.profileSectionEnd('GET_nodelist')
   })
 
+  // for testing purposes only
+  server.get('/get-checkpoints', async (_request: any, reply) => {
+    try {
+      console.log('[check-point] get-checkpoints')
```
aniketdivekar: put these logs behind verbose flag

src/API.ts
```diff
@@ -216,6 +223,28 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+      const buckets = cycleCheckpointManager.checkpointBuckets
+      console.log('[check-point] get-checkpoints buckets', buckets)
+      const checkpointData = Array.from(buckets.values()).map((bucket) => ({
```
aniketdivekar: add some comment on how much data is this, expected size of the same etc

src/API.ts
```diff
@@ -1290,6 +1319,95 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+
+      //Identify the correct manager based on the checkpoint type
+      let manager: CheckpointBucketManager<any>
+      if (checkpointType === CheckpointType.Cycle) {
+        manager = cycleCheckpointManager
+      } else if (checkpointType === CheckpointType.OriginalTx) {
+        manager = originalTxCheckpointManager
+      } else if (checkpointType === CheckpointType.Receipt) {
+        manager = receiptCheckpointManager
+      }
```
aniketdivekar: we should have this method as class method/static method in CheckpointBucketManager, something like: 
`func getCheckpointBucketManager(type: string): CheckpointBucketManager<CheckpointType.Cycle | CheckpointType.OriginalTx | CheckpointType.Receipt>`

src/API.ts
```diff
@@ -1290,6 +1319,95 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+    try {
+      const { bucketID, entries, checkpointType } = req.body
+      if (!bucketID || !entries) {
+        Logger.mainLogger.error('Invalid payload')
+        reply.status(400).send('Invalid payload')
+        return
+      }
+
+      let manager: CheckpointBucketManager<any>
+      if (checkpointType === CheckpointType.Cycle) {
```
aniketdivekar: same as above

src/Data/Collector.ts
```diff
@@ -1086,12 +1088,17 @@ export const storeReceiptData = async (
...
+    const receiptToStore = {
       ...receipt,
       receiptId: tx.txId,
       timestamp: tx.timestamp,
       applyTimestamp,
-    })
+    }
+    const receiptCheckpointData = new ReceiptCheckpointData(receipt as Receipt.Receipt)
+    // Add to checkpoint system
+    receiptCheckpointManager.addData(receiptCheckpointData, receipt.cycle.toString())
```
aniketdivekar: shouldn't this be added only after receipts are saved successfully?

src/API.ts
```diff
@@ -1290,6 +1313,127 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+        )
+        reply.status(404).send(`Bucket not found for ID=${bucketID}.`)
+        return
+      }
+
+      // Get our entries to share back
+      const ourEntries: CheckpointRadixEntry<Cycle>[] = []
+
+      // For each incoming entry, get our corresponding entry
+      for (const incomingEntry of entries) {
```
aniketdivekar: Q: shouldn't this be done after we have executed `onExchangeRadixEntries` ?

src/Data/Data.ts
```diff
@@ -1394,53 +1395,91 @@ export async function syncCyclesAndNodeListV2(
...
+          if (startCycle >= cycleToSyncTo) {
+            Logger.mainLogger.debug('Sync cycles completed!')
+            return true
+          }
+          break
+        }
+      } else {
+        Logger.mainLogger.debug(`Invalid cycle download response, attempt ${retryCount + 1} of ${MAX_RETRIES}`)
+        retryCount++
+        if (retryCount >= MAX_RETRIES) {
```
aniketdivekar, line 1492: if we have this, then do we need condition in while loop? `while (!success && retryCount < MAX_RETRIES)`

src/Data/Data.ts
```diff
@@ -1632,38 +1718,61 @@ export const syncOriginalTxs = async (): Promise<void> => {
...
+            }
+          }
+          break
+        }
+      } else {
+        Logger.mainLogger.debug(`Invalid Original-Tx download response, attempt ${retryCount + 1} of ${MAX_RETRIES}`)
+        retryCount++
+        if (retryCount >= MAX_RETRIES) {
+          Logger.mainLogger.error('Max retries reached for Original-Tx download')
+          start = end + 1
```
aniketdivekar, line 1799: we shouldn't continue right? If we failed to sync some batch of data, why should we move to next batch? 
also check other places having similar values

src/Data/Data.ts
```diff
@@ -2128,44 +2295,49 @@ export async function compareWithOldReceiptsData(lastStoredReceiptCycle = 0): Pr
...
+        }
       }
+      success = true
+      matchedCycle = downloadedCycle.counter
     }
-    success = true
-    matchedCycle = downloadedCycle.counter
+    return { success, matchedCycle }
+  } catch (error) {
+    Logger.mainLogger.error(error)
```
aniketdivekar: add some identifier text for log message
```suggestion
    Logger.mainLogger.error('compareWithOldCyclesData error: '+error)
```

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,893 @@
...
+  constructor(
+    private persistenceCallbacks: DataPersistenceCallbacks<T>,
+    checkpointType: CheckpointType
+  ) {
+    this.checkpointBuckets = new Map<string, CheckpointBucket<T>>()
+    this.validateData = persistenceCallbacks.validateData
+    this.updateData = persistenceCallbacks.updateData
+    this.checkpointType = checkpointType
+    // set to 5 minutes ago as we giveup on a bucket after 20 minutes
+    this.lastFailedBucketTime = Date.now() - 5 * 60 * 1000
```
aniketdivekar: you can simply set it to 0 when initialising. or else by default it will be 0 I think

src/API.ts
```diff
@@ -1290,6 +1313,127 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+      const result = validateRequestData(req.body, {
+        sender: 's',
+        sign: 'o',
+      })
+      if (!result.success) {
+        reply.send({ success: false, error: result.error })
+        return
+      }
+      const { bucketID, radixDigests, senderAddress, checkpointType } = req.body
+      if (!bucketID || !radixDigests) {
```
urnotsam: couldnt we just check bucketID and radixDigests in validateRequestData if those are going to be required?

urnotsam: And why not also check for senderAddress and checkpointType as well?

src/API.ts
```diff
@@ -1290,6 +1313,127 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+        Logger.mainLogger.error(
+          `Invalid payload: Missing bucketID or radixDigests in request from ${senderAddress}. CheckpointType: ${checkpointType}`
+        )
+        reply.status(400).send(`Invalid payload: Missing bucketID or radixDigests.`)
+        return
+      }
+
+      // Identify the correct manager based on the checkpoint type
+      let manager: CheckpointBucketManager<any>
```
urnotsam: simpler to do this like, but not a required change

```
const checkpointManagers: Record<CheckpointType, CheckpointBucketManager<any>> = {
  [CheckpointType.Cycle]: cycleCheckpointManager,
  [CheckpointType.OriginalTx]: originalTxCheckpointManager,
  [CheckpointType.Receipt]: receiptCheckpointManager,
};

const manager = checkpointManagers[checkpointType];
```

src/API.ts
```diff
@@ -1290,6 +1313,127 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+    try {
+      const result = validateRequestData(req.body, {
+        sender: 's',
+        sign: 'o',
+      })
+      if (!result.success) {
+        reply.send({ success: false, error: result.error })
+        return
+      }
+      const { bucketID, entries, checkpointType } = req.body
```
urnotsam, line 1372: same as above, why arent we doing this is validateRequestData?

src/API.ts
```diff
@@ -1290,6 +1313,127 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+      if (!bucketID || !entries) {
+        Logger.mainLogger.error(
+          `Invalid payload: Missing bucketID or entries in request. CheckpointType: ${checkpointType}`
+        )
+        reply.status(400).send(`Invalid payload: Missing bucketID or entries.`)
+        return
+      }
+
+      let manager: CheckpointBucketManager<any>
```
urnotsam: duplicate code from other endpoint.  refactor this for a function that gets manager

src/Config.ts
```diff
@@ -44,6 +44,7 @@ export interface Config {
     maxOriginalTxEntries: number
   }
   experimentalSnapshot: boolean
+  FAILEDBUCKETS_DIR: string
```
urnotsam: I suggest sticking to camel case convention

src/checkpoint/CheckpointData.ts
urnotsam: We should have this bucket time be a config instead of hardcoded to 5 minutes.  with that the function name below should be updated to remove the 5Mins from its name

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,893 @@
...
+  getIsLastSucceededBucketTimeOlderThan5Mins(): boolean {
+    return Date.now() - this.lastFailedBucketTime > 5 * 60 * 1000
+  }
+
+  addData(data: CheckpointData<T>, bucketID: string): void {
+    let bucket = this.checkpointBuckets.get(bucketID)
+    if (!bucket) {
+      // Determine if data.t is in milliseconds or seconds
+      let startTime = data.t
+      if (startTime > 9999999999) {
```
urnotsam, line 145: this would be more clear if we just checked` startTime.length === 10`

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,893 @@
...
+        // If greater than 10-digit Unix timestamp, assume milliseconds
+        startTime = Math.floor(startTime / 1000) // Convert to seconds
+      } else {
+        startTime = Math.floor(startTime) // Ensure it's a full second
+      }
+      const endTime = startTime + 60 // End time is 60 seconds (1 min or cycle)
```
urnotsam: since cycle length is technically configurable in our network we should also make the cycle length a config here

urnotsam (changes requested):

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,893 @@
...
+        Logger.mainLogger.error('Validation failed for data:', data)
+        return
+      }
+    }
+
+    const address = data.a.toLowerCase()
+    const radix = address.substring(0, 2)
+
+    let entry = this.radixEntries.get(radix)
+    if (!entry) {
```
aniketdivekar, line 365: this should never happen right? or this is just a safety check?

src/API.ts
jintukumardas: I think validateRequestData is a common method used by endpoints like totalData, cycleinfo, etc. I don't think those endpoints use the bucketID and radixDigests so we shouldn't add these checks in this method -
https://github.com/shardeum/archiver/blob/2373388f8520e23c21b6f852ddad75344be8cf03/src/API.ts#L1295

urnotsam: doesnt this check that the property of the object passed in is of a certain type and also throws an error if the property doesnt exist? it just goes through validateTypes which is what we were using before ajv

jintukumardas: right, sorry I misunderstood.

src/checkpoint/CheckpointData.ts
S0naliThakur: Since startTime is a number, it doesn’t have a .length property, so checking its size this way wouldn’t work. I've made a minor change for better clarity. Request you to re-reveiew

src/checkpoint/CheckpointData.ts
S0naliThakur: Changed. This is mainly for the alerting system and will likely be removed in V2, so it shouldn’t impact our system.

src/API.ts
```diff
@@ -1290,6 +1310,107 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+        )
+        reply.status(404).send(`Bucket not found for ID=${bucketID}.`)
+        return
+      }
+
+      // Get our entries to share back
+      const ourEntries: CheckpointRadixEntry<Cycle>[] = []
+
+      // For each incoming entry, get our corresponding entry
+      for (const incomingEntry of parsedEntries) {
```
aniketdivekar, line 1391: shouldn't this be done after `bucket.onExchangeRadixEntries` is done?

src/Config.ts
```diff
@@ -171,6 +181,14 @@ let config: Config = {
     MAX_CYCLES_PER_REQUEST: 100,
     MAX_BETWEEN_CYCLES_PER_REQUEST: 100,
   },
+  checkpointBucketConfig: {
+    BucketMatureAge: 11 * 60, // 11 minutes
+    cycleAge: 60, // 60 seconds
```
aniketdivekar: how much are we dependent on this config? what happens when cycle frequency is changed to a higher or lower value than 60s

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,892 @@
...
+      this.digest.itemCount = this.sortedData.length
+    }
+  }
+
+  private computeHash(): string {
+    if (!this.sortedData || this.sortedData?.length === 0) {
+      return Crypto.hash('').toLowerCase()
+    }
+
+    const hashToCompute = this.sortedData
```
aniketdivekar, line 94: in `updateDigest` we are explicitly sorting this.sortedData and then calculating `this.computeHash()`. 
We don't need to sort it here?

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,892 @@
...
+  constructor(
+    private persistenceCallbacks: DataPersistenceCallbacks<T>,
+    checkpointType: CheckpointType
+  ) {
+    this.checkpointBuckets = new Map<string, CheckpointBucket<T>>()
+    this.validateData = persistenceCallbacks.validateData
+    this.updateData = persistenceCallbacks.updateData
+    this.checkpointType = checkpointType
+    // set to 5 minutes ago as we giveup on a bucket after 20 minutes
+    this.lastFailedBucketTime = Date.now() - config.checkpointBucketConfig.lastFailedBucketDuration
```
aniketdivekar: this can be set to 0 when initialising

src/checkpoint/CheckpointData.ts
```diff
@@ -0,0 +1,892 @@
...
+        // Only request repair if:
+        // 1. We have a majority hash
+        // 2. The majority hash has enough votes
+        // 3. Our hash is different from majority hash
+        // 4. Our hash has fewer votes than the majority hash
+        if (
+          majorityHash &&
+          maxVotes >= majorityThreshold &&
+          majorityHash !== localEntry.digest.hash &&
+          (tally.hashTally.get(localHash) || 0) < maxVotes
```
aniketdivekar: 3rd and 4th condition should be OR right? either of them is true meaning we need to repair
```
majorityHash &&
maxVotes >= majorityThreshold &&
(majorityHash !== localEntry.digest.hash || (tally.hashTally.get(localHash) || 0) < maxVotes)
```
or only one of them is needed 
is it possible that local hash is matching majority hash but then doesn't have enough votes? If that's the case, then it must be business logic issue right?

src/checkpoint/CycleData.ts
```diff
@@ -0,0 +1,168 @@
...
+//Represents a single piece of cycle data
+export class CycleCheckpointData extends CheckpointData<Cycle> {
+  constructor(cycle: Cycle) {
+    const cycleHash = Crypto.hash(StringUtils.safeStringify(cycle)).toLowerCase()
+
+    super(
+      cycleHash.substring(0, 2), // address (first 2 chars)
+      cycle.cycleRecord.start, // timestamp from cycleRecord
+      cycleHash, // hash
+      0, // class type 0 for cycle
```
aniketdivekar: I think we can use `CheckpointType.Cycle.value` instead of hardcoded 0

src/checkpoint/OriginalTxsData.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class OriginalTxCheckpointData extends CheckpointData<OriginalTxData> {
+  constructor(data: OriginalTxData) {
+    const originalTxHash = Crypto.hash(StringUtils.safeStringify(data)).toLowerCase()
+
+    super(
+      originalTxHash.substring(0, 2), // address (first 2 chars)
+      data.timestamp, // timestamp
+      originalTxHash, // hash
+      1, // class_type 1 for originalTx
```
aniketdivekar: `CheckpointType.OriginalTx.value`

src/checkpoint/ReceiptData.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export class ReceiptCheckpointData extends CheckpointData<ReceiptType> {
+  constructor(receipt: ReceiptType) {
+    const receiptHash = Crypto.hash(StringUtils.safeStringify(receipt)).toLowerCase()
+
+    super(
+      receiptHash.substring(0, 2), // address (first 2 chars)
+      receipt.tx.timestamp, // timestamp
+      receiptHash, // hash
+      2, // class type 2 for receipts
```
aniketdivekar: `CheckpointType.Receipt.value`

src/checkpoint/ReceiptData.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const appValidation = await verifyAppReceiptData(data.d, null, [], [])
+    return appValidation.valid
+  } catch (err) {
+    Logger.mainLogger.error('ValidateData failed:', err)
+    return false
+  }
+}
+
+async function updateData(data: CheckpointData<ReceiptType>): Promise<void> {
```
aniketdivekar, line 113: I wonder why we have added queries here.. can we not invoke existing methods to insert (insertReceipt or bulkInsertReceipts)
(same for cycles and originaTxs)

src/dbstore/cycles.ts
```diff
@@ -77,6 +87,10 @@ export async function bulkInsertCycles(cycles: Cycle[]): Promise<void> {
 
 export async function updateCycle(marker: string, cycle: Cycle): Promise<void> {
   try {
+    // Create a checkpoint before updating
+    const checkpointData = new CycleCheckpointData(cycle)
+    const bucketID = calculateBucketID(cycle)
+    cycleCheckpointManager.addData(checkpointData, bucketID)
```
aniketdivekar: .addData or .update?

src/server.ts
```diff
@@ -177,9 +180,29 @@
...
+        cycleCheckpointManager.update()
+        receiptCheckpointManager.update()
+        originalTxCheckpointManager.update()
+      } catch (err) {
+        Logger.mainLogger.error('Error updating checkpoints:', err)
+      }
+    } else {
+      Logger.mainLogger.error('One or more checkpoint managers not initialized')
+    }
+  }, config.checkpointUpdateInterval)
```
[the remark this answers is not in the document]
aniketdivekar: we had discussed problem when using `setInterval`. 1min probably isn't enough where it has to sync lot of data. I suggest we simply have an infinite blocking loop.

src/API.ts
S0naliThakur: This has been removed during clean up

src/API.ts
S0naliThakur: This is available in the latest commit

src/Data/Collector.ts
S0naliThakur: It is okay to include the checkpoint here as even if database writing ends up being unsuccessful, the majority consensus in the checkpoint system will fix the the CheckpointData entry and write the correct data to the database

src/checkpoint/ReceiptData.ts
S0naliThakur: We modified the insert methods to create the checkpoint data entry while writing to the database, ensuring we don’t miss any incoming data. Hence we are not resuing the insert methods from dbstore

src/dbstore/cycles.ts
S0naliThakur: addData only to include the data checkpointData object into the appropriate bucket

src/server.ts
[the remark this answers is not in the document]
S0naliThakur: Added an independent method to handle writing and not blocking the flow. one minute `setInterval` duration will not create any issue anymore

src/checkpoint/ReceiptData.ts
aniketdivekar: I think this is going to create issue later... we have to maintain insert functions at two places. in existing flow, we can have conditional checkpoint entry.

src/API.ts
S0naliThakur: This is the data we received post consensus that needs to be included in our local data. `bucket.onExchangeRadixEntries` will perform the merging part

src/Config.ts
S0naliThakur: Not a very huge dependency as most of the operations depend on the start time of a cycle. The cycleAge `config` is used to obtain the end time of a bucket

src/checkpoint/CheckpointData.ts
S0naliThakur: Not needed as we use it only after sorting the data to compute hash for maintaining the hash consistency

src/checkpoint/CheckpointData.ts
S0naliThakur: This is meant to be used for alerting but removing it in v2 altogether. Can we revisit this in v2?

src/checkpoint/CheckpointData.ts
S0naliThakur: correct. Just a safety check

src/API.ts
S0naliThakur: As mentioned previously, this is the data we received post consensus that needs to be included in our local data. onExchangeRadixEntries will perform the merging part

src/Data/Data.ts
S0naliThakur: this helps in ensuring that we receive a response before proceeding with the next set of data. `success` and `MAX_RETRY` together prevent the infinite loop problem

src/checkpoint/ReceiptData.ts
S0naliThakur: fixed in the new commit

src/Data/Data.ts
S0naliThakur: removed

src/Data/Data.ts
S0naliThakur: added

src/checkpoint/CheckpointData.ts
S0naliThakur: Since both majorityHash and maxVotes are derived in the same loop, we can remove the fourth condition. Removed in the latest commit

src/checkpoint/CycleData.ts
S0naliThakur: done

src/Data/Data.ts
S0naliThakur: as answered in another thread, both are needed to ensure we dont go into infitnite loop

src/checkpoint/CheckpointData.ts
S0naliThakur: will be removed in v2

urnotsam (dismissed):

aniketdivekar (dismissed):

urnotsam (dismissed):

src/API.ts
```diff
@@ -1290,6 +1310,107 @@ export function registerRoutes(server: FastifyInstance<Server, IncomingMessage,
...
+      Logger.mainLogger.error(
+        `Error processing shareCheckpointRadixDigests for checkpoint type ${req.body.checkpointType}: ${err.message}`
+      )
+      reply.status(500).send('Server error')
+    }
+  })
+
+  server.post('/exchangeCheckpointRadixEntries', async (req: any, reply) => {
+    try {
+      const result = validateRequestData(req.body, {
```
aniketdivekar, line 1361: * validate if sender is in whitelisted archiver
* also need crypto.verify if not happening already in `validateRequestData`

S0naliThakur: `validateRequestData` does both

src/dbstore/originalTxsData.ts
```diff
@@ -106,7 +119,7 @@ export async function queryOriginalTxDataCount(startCycle?: number, endCycle?: n
     }
     originalTxsData = await db.get(originalTxDataDatabase, sql, values)
   } catch (e) {
-    console.log(e)
+    Logger.mainLogger.error(e)
```
aniketdivekar, line 132:
```suggestion
    Logger.mainLogger.error('queryOriginalTxDataCount ', e)
```

aniketdivekar (dismissed):

urnotsam (dismissed):

aniketdivekar (dismissed):

urnotsam (dismissed):

devendra-shardeum:
please remove package-lock.json from the pr

aniketdivekar (dismissed):

aniketdivekar (dismissed):

urnotsam (dismissed):

urnotsam (dismissed):

urnotsam (dismissed):

aniketdivekar (approved):

urnotsam (approved):

src/Data/Data.ts
```diff
@@ -1674,38 +1757,60 @@ export const syncOriginalTxs = async (): Promise<void> => {
...
+          start: start,
+          end: end,
+        },
+        QUERY_TIMEOUT_MAX
+      )
+
+      if (res && res.originalTxs) {
+        const downloadedOriginalTxs = res.originalTxs
+        Logger.mainLogger.debug('Downloaded Original-Txs: ', downloadedOriginalTxs.length)
+        await storeOriginalTxData(downloadedOriginalTxs)
```
devendra-shardeum, line 1778: all the queryFromArchivers call and their response aren't going through crypto.verify and that's needed to verify that the data is coming from our own allowed archivers and this is for all types of data receipts, cycles, accounts and transactions.. I see that the data has been signed on the api endpoint and should be verified at the receiver side, what did i miss ?  @S0naliThakur cc : @aniketdivekar

S0naliThakur: As discussed, since requests are sent only to the allowed archivers through this method, we can be certain that the response comes from an authorized source. However, to align with security guidelines, a PR has been created to include signature verification [here](https://github.com/shardeum/archiver/pull/162).

PR #135, opened 2025-02-05
Shard 1244

devendra-shardeum:
* smoke testing done
* also tested ChangeNetworkParam, ChangeConfig types of txs from lt2


src/services/transactionVerification.ts
```diff
@@ -0,0 +1,366 @@
...
+        if ( !isValid ) {
+          Logger.mainLogger.info(`Single signed tx failed ${JSON.stringify(tx)}`)
+          return { result: 'fail', reason: 'Invalid Signature' }
+        }
+
+        return { result: 'pass', reason: 'valid' }
+      }
+    }
+
+    if (tx?.isDebugTx) {
```
abdulazeem-tk4vr: just nitpicking since a similar format is followed on the shardeum/server side and I have used the same for SEC-526.
Please use this, thanks!

// eslint-disable-next-line @typescript-eslint/no-explicit-any
export function isDebugTx(tx: any): boolean {
  return tx.isDebugTx != null
}

src/services/transactionVerification.ts
```diff
@@ -0,0 +1,366 @@
...
+    internalTx.internalTXType === InternalTXType.ApplyNetworkParam
+  )
+}
+
+export function isInternalTx(timestampedTx: any ): boolean {
+
+    if(!timestampedTx) return false;
+
+    return (
+      !timestampedTx.raw &&
```
abdulazeem-tk4vr: why are we allowing the raw portion to be null?

Would suggest a copy paste from the validator side for convenience 

// eslint-disable-next-line @typescript-eslint/no-explicit-any
export function isInternalTx(timestampedTx: any): boolean {
  if (timestampedTx && timestampedTx.raw) return false
  if (timestampedTx && timestampedTx.isInternalTx) return true
  if (timestampedTx && timestampedTx.tx && timestampedTx.tx.isInternalTx) return true
  return false
}

src/Utils.ts
```diff
@@ -568,3 +571,51 @@ export function createDirectories(pathname: string): void {
   pathname = pathname.replace(/^\.*\/|\/?[^/]+\.[a-z]+|\/$/g, '') // Remove leading directory markers, and remove ending /file-name.extension
   fs.mkdirSync(path.resolve(__dirname, pathname), { recursive: true }) // eslint-disable-line security/detect-non-literal-fs-filename
 }
+
+
+export function verifyMultiSigs(
```
abdulazeem-tk4vr, line 576: this seems to have been taken from the validator side, just a heads-up for the reviewers.

abdulazeem-tk4vr: nothing concerning here

src/server.ts
```diff
@@ -180,6 +182,11 @@ async function start(): Promise<void> {
     Logger.mainLogger.debug('We are not first archiver. Syncing and starting archive-server')
     syncAndStartServer()
   }
+
+  setTimeout(() => {
+    scheduleMultiSigKeysSyncFromNetConfig();
+  }, 60 * 1000); // Start after 60 seconds
```
abdulazeem-tk4vr, line 188: can we make this configurable? I'm sorry for nitpicking but I get the same comment on every magic number I add.

src/services/transactionVerification.ts
```diff
@@ -0,0 +1,366 @@
...
+
+type Response = {
+  result: string
+  reason: string
+}
+
+
+export const scheduleMultiSigKeysSyncFromNetConfig =  (): void => {
+
+  setInterval(async () => {
```
abdulazeem-tk4vr, line 58: I am a bit confused about why we need 2 schedulers. Wouldn't scheduling the parent function and just calling the child function in it suffice?

src/services/transactionVerification.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  } catch (error) {
+    Logger.mainLogger.error('error in syncKeysFromNetworkConfig', error)
+  }
+
+}
+
+
+
+export const verifyTransaction = (tx: any) : Response  => {
```
abdulazeem-tk4vr, line 130: again heads up for new reviewers, this has been taken from `validateTransaction` of shardeum/src/setup

abdulazeem-tk4vr (changes requested):

src/services/transactionVerification.ts
devendra-shardeum: one is scheduler .. other just delays for a cycle to get it started .. it's setTimeout

src/services/transactionVerification.ts
devendra-shardeum: i think this is also faulty and mine as well.. in yours it will let pass tx.isDebugTx = false as well .. better implementation imo is : 

if ( tx?.isDebugTx === true ) {
// logic to let it pass 
}

src/services/transactionVerification.ts
devendra-shardeum: image

taken from validators side

devendra-shardeum: noted

src/server.ts
devendra-shardeum: other schedulers in archiver .. like collecting missing receipts/txs are also not pulled out to be configured from config.ts : 

image


upto you, please tell me if this needs to go there

devendra-shardeum: but adding mine in config.. i also agree it can and should be configured via configuration

abdulazeem-tk4vr (approved):

urnotsam (approved):

PR #139, opened 2025-02-12
SHARD-1807: empty archivers list in archiver-config.json + use absolute path to l…

ahmxdiqbal:
…oad archiver-config.json


archiver-config.json
```diff
@@ -94,7 +43,7 @@
       "0x1e5e12568b7103E8B22cd680A6fa6256DD66ED76": 5,
       "0xa58169308e7153B5Ce4ca5cA515cC4d0cBE7770B": 5
     },
-    "minSigRequired": 1,
```
urnotsam: was this an intentional change?

ahmxdiqbal: no, thanks for catching that

urnotsam (approved):

PR #140, opened 2025-02-18
SHARD-1875: BB: use most common cycle marker

Glitch18:
https://linear.app/shm/issue/SHARD-1875/[bb]-39626-malicious-validator-can-overwrite-any-cycle-data

### DIFF
--- a/src/Data/Data.ts
+++ b/src/Data/Data.ts
@@ -191,17 +191,17 @@ export async function unsubscribeDataSender(
 
 export function initSocketClient(node: NodeList.ConsensusNodeInfo): void {
   if (config.VERBOSE) Logger.mainLogger.debug('Node Info to socket connect', node)
-  
+
   try {
     const socketClient = ioclient.connect(`http://${node.ip}:${node.port}`, {
       query: {
         data: JSON.stringify(
-        Crypto.sign({
-          publicKey: State.getNodeInfo().publicKey, 
-          timestamp: Date.now(),
-          intendedConsensor: node.publicKey
-       })
-      ),
+          Crypto.sign({
+            publicKey: State.getNodeInfo().publicKey,
+            timestamp: Date.now(),
+            intendedConsensor: node.publicKey
+          })
+        ),
       },
     })
     socketClients.set(node.publicKey, socketClient)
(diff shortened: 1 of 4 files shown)


src/Data/Data.ts
```diff
@@ -387,27 +387,39 @@ export function collectCycleData(
...
-      if (value['receivedTimes'] >= minCycleConfirmations) {
-        cycleToSave.push(cycle)
-        value['saved'] = true
+      continue
+    }
+
+    // Check if any of the markers for this cycle have received the minimum number of confirmations
+    if (Object.values(receivedCycleTracker[cycle.counter]).some((value) => value['receivedTimes'] >= minCycleConfirmations)) {
+      // Find cycle with max receivedTimes
+      const maxReceivedTimes = Math.max(
```
aniketdivekar, line 409: I am not sure if the ask of using most common cycle marker is addressed here.

aniketdivekar: spoke on DM. looks good

Glitch18: It is. We find the cycle marker received most number of times and use it.

aniketdivekar:
PR looks okay to me. Waiting for details on dev testing

Glitch18:
Dev testing details:

Devnet-us was started with two archivers, one with the dev branch while other with these changes. We compared the last 100 cycles across both archviers and things look good.

urnotsam (dismissed):

urnotsam (approved):

abdulazeem-tk4vr (approved):

src/Data/Data.ts
```diff
@@ -387,27 +387,39 @@ export function collectCycleData(
...
+        receivedCycleTracker[cycle.counter][cycleToSave[i].marker]['saved'] = false
       }
-      if (value['receivedTimes'] >= minCycleConfirmations) {
-        cycleToSave.push(cycle)
-        value['saved'] = true
+      continue
+    }
+
+    // Check if any of the markers for this cycle have received the minimum number of confirmations
+    if (Object.values(receivedCycleTracker[cycle.counter]).some((value) => value['receivedTimes'] >= minCycleConfirmations)) {
```
PudgyPug, line 407: we could save some loop iteration when doing something like this no?
```
const cycleValues = Object.values(receivedCycleTracker[cycle.counter]);

let maxReceivedTimes = 0;
let markerValue: typeof cycleValues[number] | null = null;

// Single loop to find maxReceivedTimes and the first marker with that value
for (const value of cycleValues) {
  if (value.receivedTimes >= minCycleConfirmations) {
    if (value.receivedTimes > maxReceivedTimes) {
      maxReceivedTimes = value.receivedTimes;
      markerValue = value;
    }
  }
}

if (markerValue) {
  cycleToSave.push(markerValue.cycleInfo);
  markerValue.saved = true;
}
```

PudgyPug: nvm just got told me this is called once cycle anyway

PR #143, opened 2025-02-25
SHARD-1892 : execution-shard-verification-in-archiver

abdulazeem-tk4vr:
### 🚀 Summary

This PR introduces two key improvements to the archiver’s receipt handling and validation mechanisms:

1. **Authorization & verification of signatures** for both **Global** and **Non-Global** transaction receipts  
2. **Coupling of `OTxData` storage with receipt validation**
3. **Addressing Scalability issues with RobustQuery and relying solely on SignedReceipt88

---

### 📌 Context  
Refer to [SHARD-1892](https://linear.app/shm/issue/SHARD-1892/execution-shard-verification-in-archiver) for the original issue description.

---

### ✅ What was added

- Integrated **authorization logic** directly into the **receipt signature verification flow**, eliminating a previously independent and exploitable separation of logic.
- Fixed **majority signature calculation**, which previously allowed a single signature to pass verification in a pack of 100.
- Prevented **duplicate signature verification** efforts across modules.
- Ensured **correct payloads are used** during verification to avoid data store failures.
- Coupled **OTxData verification** with receipt validation to prevent tampering with transactional metadata.

---

### 🔒 Why it was needed

- In high-load situations, contacting multiple nodes will lead to a delay in receipt verification. This becomes a bottleneck.
- Votes from unauthorized shards could previously be accepted due to `shardKey` manipulation.
- Lack of coupling between authorization and signature verification created security loopholes.
- Invalid payload usage led to failed receipt storage.
- `OTxData` was not being verified, exposing the system to silent tampering.
- See [SEC-526](#sec-526) and [SHARD-1882](#shard-1882) for related vulnerability context.

---

### 🧵 Follow-ups

A new issue has been raised concerning improper verification of `beforeStateHashes` and `afterStateHashes` before insertion into the DB. This could result in unauthorized **state transitions** and **balance manipulation**.

The security team has labeled this as a **critical vulnerability**, and it will be addressed in a separate follow-up ticket.

---

#### SEC-526

> When the archiver receives a receipt from a validator, the `executionShardKey` in the receipt is not correctly tied to the account IDs being updated.  
>  
> If an attacker controls a single shard, they could submit valid-looking signatures for a transaction that updates accounts outside their shard.  
>  
> **Impact**: Critical  
> **Context**: `archive-server` repo – `storeReceiptData`, `verifyReceiptData`

---

#### SHARD-1882

> The archiver currently **trusts the validator's original transaction data** without verifying it against consensus.  
>  
> Fix: Receipt flow should bundle both the receipt and the original transaction. Once the receipt is verified via `txId`, the corresponding transaction can be safely persisted.  
>  
> Requires:
> - Payload validation  
> - Signature verification  
> - Proper authorization


src/Data/Collector.ts
```diff
@@ -63,6 +65,68 @@ export interface ReceiptVerificationResult {
...
+    if (!cycleShardData.parititionShardDataMap.get(homePartition).coveredBy[node.id]) {
+      Logger.mainLogger.error(
+        `The node with public key ${nodePubKey} of the receipt ${txId} with ${timestamp} is not in the execution group of the tx`
+      )
+      if (nestedCountersInstance)
+        nestedCountersInstance.countEvent('receipt', 'node_not_in_execution_group_of_tx')
+      continue
+    }
+
+    acceptableSigners.add([index, sign])
```
aniketdivekar: is `node.publicKey` as good as unique signers? if so, we should lowercase and calculate uniqueness. There have been some reports that same value but different case (e.g. 'abcd', 'Abcd', 'aBcd') are considered as different but they are actually the same.

src/Data/Collector.ts
```diff
@@ -63,399 +62,308 @@ export interface ReceiptVerificationResult {
...
-          'globalModification_sign_node_not_in_execution_group_of_tx'
-        )
-      continue
-    }
-
-    uniqueSigners.add(nodePubKey)
-  }
-  isReceiptMajority = (uniqueSigners.size / votingGroupCount) * 100 >= config.requiredMajorityVotesPercentage
+  const voteHash = calculateVoteHash(proposal)
+  const acceptableSigners = fetchAuthorizedSigners(
```
devendra-shardeum, line 317: fetchAuthorizedSigners returns acceptableSigners which will contain signs and not public key.. for same receipt we can generate multiple signs for the same public key and those will all be valid .. it'll break the flow in goodSignatures counting below where we are relying on the public key as the key and will be unique but again that's not lower cased and not also length not checked..

urnotsam (dismissed):
I wont hold the PR up over this, but we should really not be leaving dead / comment out code in.  We can clean it up later on this one.

urnotsam (approved):

src/Data/Collector.ts
mssabr01: acceptableSigners now contains public keys

mssabr01 (approved):

PR #148, opened 2025-02-28
SHARD-1876: Archiver Hardening V2+V3

S0naliThakur:
Linear: https://linear.app/shm/issue/SHARD-1876
Summary: Archiver hardening v2+v3

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -22,6 +22,7 @@ export interface Config {
     originalTxDataDB: string
     processedTxDB: string
     txDigestDB: string
+    checkpointStatusDB: string
   }
   DATASENDER_TIMEOUT: number
   RATE_LIMIT: number // number of allowed request per second,
@@ -148,6 +149,7 @@ let config: Config = {
     originalTxDataDB: 'originalTxsData.sqlite3',
     processedTxDB: 'processedTransactions.sqlite3',
     txDigestDB: 'txDigest.sqlite3',
+    checkpointStatusDB: 'checkpointStatus.sqlite3',
   },
   DATASENDER_TIMEOUT: 1000 * 60 * 5,
   RATE_LIMIT: 100, // 100 req per second,
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -193,13 +193,13 @@ let config: Config = {
   },
   checkpoint: {
     bucketConfig: {
-      BucketMatureAge: 3 * 60, // 11 minutes
+      BucketMatureAge: 11 * 60, // 11 minutes
       cycleAge: 60, // 60 seconds
-      GiveUpAge: 6 * 60, // 20 minutes
+      GiveUpAge: 20 * 60, // 20 minutes
       lastFailedBucketDuration: 5 * 60 * 1000, // 5 minutes
       RadixDepth: 2, // 2 nibbles (1 hex char)
-      allowCheckpointUpdates: true,
-      allowCheckpointStorage: true
+      allowCheckpointUpdates: false,
+      allowCheckpointStorage: false
     },
     updateInterval: 60 * 1000, // 1 minute in milliseconds  in milliseconds
     syncInterval: 10000, // 10 seconds in milliseconds
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -36,16 +36,21 @@ export interface Config {
   }
   ARCHIVER_MODE: string
   DevPublicKey: string
-  checkpointBucketConfig: {
-    BucketMatureAge: number //  start sharing hashes after this age is reached.
-    cycleAge: number // In second
-    GiveUpAge: number // eventual give up age.  write bucket to disk in this case and raise warnings/alerts
-    lastFailedBucketDuration: number // in milliseconds, used for alerting
-    RadixDepth: number // 16 way trie depth in nibbles (one hex char)
-    allowCheckpointUpdates: boolean
-    allowCheckpointStorage: boolean
+  checkpoint: {
+    bucketConfig: {
+      BucketMatureAge: number
+      cycleAge: number
+      GiveUpAge: number
+      lastFailedBucketDuration: number
+      RadixDepth: number
+      allowCheckpointUpdates: boolean
+      allowCheckpointStorage: boolean
+    }
+    updateInterval: number
+    syncInterval: number
+    maxCyclesToSync: number
+    syncOnStartup: boolean
   }
-  checkpointUpdateInterval: number // 1 minute in milliseconds
   dataLogWrite: boolean
   dataLogWriter: {
     dirName: string
@@ -124,12 +129,6 @@ export interface Config {
   }
   maxRecordsPerRequest: number // this is the equiavlent of the accountBucketSize config variable used by the validators to fetch records from the archiver
   multisigKeysSyncFromNetworkInternal: number // in seconds
-  checkpointV2: {
-    enabled: boolean
-    syncInterval: number
-    maxCyclesToSync: number
-    syncOnStartup: boolean
-  }
   minCycleConfirmationsToSave: number // this is the minimum numbers of nodes that we need to a see a cycle from to save it
 }
 
@@ -192,18 +191,17 @@ let config: Config = {
     MAX_CYCLES_PER_REQUEST: 100,
     MAX_BETWEEN_CYCLES_PER_REQUEST: 100,
   },
-  checkpointBucketConfig: {
-    BucketMatureAge: 11 * 60, // 11 minutes
-    cycleAge: 60, // 60 seconds
-    GiveUpAge: 20 * 60, // 20 minutes
-    lastFailedBucketDuration: 5 * 60 * 1000, // 5 minutes
-    RadixDepth: 2, // 2 nibbles (1 hex char)
-    allowCheckpointUpdates: false,
-    allowCheckpointStorage: false
-  },
-  checkpointUpdateInterval: 60 * 1000, // 1 minute in milliseconds  in milliseconds
-  checkpointV2: {
-    enabled: false, // Enable V2 checkpoint system
+  checkpoint: {
+    bucketConfig: {
+      BucketMatureAge: 3 * 60, // 11 minutes
+      cycleAge: 60, // 60 seconds
+      GiveUpAge: 6 * 60, // 20 minutes
+      lastFailedBucketDuration: 5 * 60 * 1000, // 5 minutes
+      RadixDepth: 2, // 2 nibbles (1 hex char)
+      allowCheckpointUpdates: true,
+      allowCheckpointStorage: true
+    },
+    updateInterval: 60 * 1000, // 1 minute in milliseconds  in milliseconds
     syncInterval: 10000, // 10 seconds in milliseconds
     maxCyclesToSync: 100, // Maximum number of cycles to sync in one go
     syncOnStartup: false, // Sync missing checkpoints on startup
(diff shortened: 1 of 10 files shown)


src/Data/Cycles.ts
```diff
@@ -424,8 +424,17 @@ export async function getNewestCycleFromArchivers(): Promise<P2PTypes.CycleCreat
   ): Promise<P2PTypes.CycleCreatorTypes.CycleData[]> => {
     const response = (await postJson(
       `http://${node.ip}:${node.port}/cycleinfo`,
-      data
+      Crypto.sign(data)
```
aniketdivekar, line 427: this is not needed.. already happening on [here](https://github.com/shardeum/archiver/pull/148/files#diff-84e4961e1638afd7078fdc5457fda20627ba9707034ac5c473e60916bffd34baR420)

src/Data/Data.ts
```diff
@@ -2407,3 +2407,51 @@ async function downloadOldCycles(
     endCycle = startCycle - 1
   }
 }
+
+/**
+ * Syncs cycle data for a specific cycle
+ * @param cycle The cycle number to sync
+ * @returns True if successful, false otherwise
+ */
+export async function syncCycleData(cycle: number): Promise<boolean> {
```
aniketdivekar, line 2432: why do we need a new method? we should have existing method to sync cycles (maybe between range of something)

src/Config.ts
```diff
@@ -123,6 +123,12 @@ export interface Config {
   }
   maxRecordsPerRequest: number // this is the equiavlent of the accountBucketSize config variable used by the validators to fetch records from the archiver
   multisigKeysSyncFromNetworkInternal: number // in seconds
+  checkpointV2: {
```
aniketdivekar: not sure why are we calling v2?.. its only one solution to the problem being implemented in multiple phases. let's not introduce confusion with calling this v2.

src/Config.ts
```diff
@@ -123,6 +123,12 @@ export interface Config {
...
+    enabled: boolean
```
aniketdivekar: Also, we don't need separate flag here... checkpoint as a whole is enabled or disabled.

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,283 @@
...
+        let radixesWithMajority = 0
+
+        for (const [radix, entry] of bucket.peerRadixDigests.entries()) {
+            let maxVotes = 0
+            let totalVotes = 0
+
+            // Find the hash with the most votes
+            for (const [hash, votes] of entry.hashTally.entries()) {
+                totalVotes += votes
+                if (votes > maxVotes) {
```
aniketdivekar: should this be ? If not, then `totalVotes` is not needed as its not used anywhere else 
```suggestion
                if (totalVotes > maxVotes) {
```

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        // Calculate the percentage of archivers that agree with us
+        // This is a rough estimate based on the number of radixes with majority consensus
+        const consensusPercentage = bucket.radixEntries.size > 0
+            ? radixesWithMajority / bucket.radixEntries.size
+            : 0
+
+        // Estimate the number of archivers that agree with us
+        matchedArchivers = Math.floor(consensusPercentage * totalArchivers)
```
aniketdivekar: can this be simply `radixesWithMajority` ? I am just worried if this calculation can give us 2 but it was actually 3 because of some magic value of `consensusPercentage`

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,283 @@
...
+        const status = matchedArchivers >= requiredMajority
+            ? CheckpointSyncStatus.COMPLETED
+            : CheckpointSyncStatus.FAILED
+
+        // Record the checkpoint status
+        const checkpointStatus: CheckpointStatus = {
+            cycle,
+            type: mapCheckpointTypeToStatusType(checkpointType),
+            status,
+            timestamp: Date.now(),
```
aniketdivekar: we can call this column as `created_at`

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,305 @@
...
+
+            // Check if we need to sync original tx data
+            const originalTxStatus = await getCheckpointStatus(cycle, CheckpointStatusType.ORIGINAL_TX)
+            if (!originalTxStatus || originalTxStatus.status === CheckpointSyncStatus.PENDING || originalTxStatus.status === CheckpointSyncStatus.FAILED) {
+                await updateCheckpointStatus(cycle, CheckpointStatusType.ORIGINAL_TX, CheckpointSyncStatus.SYNCING)
+
+                try {
+                    // Sync original tx data
+                    Logger.mainLogger.info(`Syncing original tx data for cycle ${cycle}`)
+                    await Data.syncOriginalTxsByCycle(cycle, cycle)
```
aniketdivekar: how are we choosing the archiver to fetch this data from? if we are choosing archiver which doesn't have correct data, then how does it impact this syncing?

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,305 @@
...
+
+                    // Mark as completed
+                    await updateCheckpointStatus(cycle, CheckpointStatusType.ORIGINAL_TX, CheckpointSyncStatus.COMPLETED)
```
aniketdivekar: this shouldn't be marked completed here I think. This should be marked completed where we know our hash has majority ... what if `await Data.syncOriginalTxsByCycle(cycle, cycle)` still doesn't sycn everything correctly as we may be choosing a random archiver to sync from? 

(This is applicable to cycles, receipts also)

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,305 @@
...
+            }
+
+            // Otherwise, continue with the sync
+            await syncMissingCheckpoints()
+        } catch (error) {
+            Logger.mainLogger.error('Error in scheduled checkpoint sync:', error)
+        }
+
+        // Schedule the next sync if still active
+        if (syncActive) {
```
aniketdivekar: let's say we caught up with the network and have latest cycle with correct data.. what if something goes wrong in between... if we stop triggering `syncCheckpoints` in a running archiver, we'll never be able to recover w/o a restart right?

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,305 @@
...
+import { cycleCheckpointManager } from './CycleData'
+import { receiptCheckpointManager } from './ReceiptData'
+import { originalTxCheckpointManager } from './OriginalTxsData'
+
+/**
+ * Converts CheckpointType to CheckpointStatusType
+ * @param checkpointType The CheckpointType to convert
+ * @returns The corresponding CheckpointStatusType
+ */
+function mapCheckpointTypeToStatusType(checkpointType: CheckpointType): CheckpointStatusType {
```
aniketdivekar: this can be a map constant... we don't need a method for this

src/checkpoint/CheckpointV2.ts
```diff
@@ -0,0 +1,305 @@
...
+        for (const [id, bucket] of this.checkpointBuckets.entries()) {
+            if (currentTime > bucket.GiveUpAge) {
+                // Record the checkpoint status before removing
+                await recordCheckpointStatus(bucket, this.checkpointType)
+                toRemove.push(id)
+            }
+        }
+
+        // Call the original update method
+        await originalUpdate.call(this)
```
aniketdivekar: let's call this `originalCycleUpdate`? similar to `originalReceiptUpdate`/`originalOriginalTxUpdate`

src/dbstore/checkpointStatus.ts
```diff
@@ -0,0 +1,344 @@
...
+}
+
+/**
+ * Inserts or updates a checkpoint status record
+ * @param status The checkpoint status to insert or update
+ */
+export async function upsertCheckpointStatus(status: CheckpointStatus): Promise<void> {
+  try {
+    const sql = `
+      INSERT OR REPLACE INTO checkpoint_status
```
aniketdivekar, line 42: `INSERT OR REPLACE INTO` has meaning only if there is a unique constraint. If not, in race conditions, you may cause duplicate records.  Hope its taken care of

src/dbstore/checkpointStatus.ts
```diff
@@ -0,0 +1,344 @@
...
+/**
+ * Gets all checkpoint statuses with a specific status
+ * @param status The status to filter by
+ * @returns Array of checkpoint statuses
+ */
+export async function getCheckpointStatusesByStatus(
+  status: CheckpointSyncStatus
+): Promise<CheckpointStatus[]> {
+  try {
+    const sql = `
```
aniketdivekar, line 315: this query should be paginated .. if I have million records in this table then, it will try to return million records... basically will go OOM

src/dbstore/index.ts
```diff
@@ -108,6 +113,16 @@ export const initializeDB = async (config: Config): Promise<void> => {
     processedTxDatabase,
     'CREATE INDEX if not exists `processedTxs_cycle_idx` ON `processedTxs` (`cycle`)'
   )
+
+  // Create checkpoint_status table
+  await runCreate(
+    checkpointStatusDatabase,
+    'CREATE TABLE if not exists `checkpoint_status` (`cycle` NUMBER NOT NULL, `type` TEXT NOT NULL, `status` TEXT NOT NULL, `timestamp` BIGINT NOT NULL, `totalArchivers` NUMBER NOT NULL, `matchedArchivers` NUMBER NOT NULL, `failedRadixes` JSON, `lastSyncAttempt` BIGINT, `syncAttempts` NUMBER DEFAULT 0, PRIMARY KEY (`cycle`, `type`))'
```
aniketdivekar:
```suggestion
    'CREATE TABLE if not exists `checkpoint_status` (`cycle` NUMBER NOT NULL, `type` TEXT NOT NULL, `status` TEXT NOT NULL, `created_at` BIGINT NOT NULL, `totalArchivers` NUMBER NOT NULL, `matchedArchivers` NUMBER NOT NULL, `failedRadixes` JSON, `lastSyncAttempt` BIGINT, `syncAttempts` NUMBER DEFAULT 0, PRIMARY KEY (`cycle`, `type`))'
```

src/server.ts
```diff
@@ -307,15 +326,83 @@ async function syncAndStartServer(): Promise<void> {
...
     if (!cycleResult.success) {
-      throw Error(
-        'The last saved 10 cycles data does not match with the archiver data! Clear the DB and start the server again!'
-      )
-    }
+      // Get the latest cycle from archivers to know how far we need to sync
+      const latestNetworkCycle = await Cycles.getNewestCycleFromArchivers()
+
+      // Find the last valid cycle we have
+      let lastValidCycle = cycleResult.matchedCycle
```
aniketdivekar, line 339: we should extract method for this logic

src/Data/Cycles.ts
S0naliThakur: Without signing here we were getting signature verification errors in a few places

src/Data/Data.ts
S0naliThakur: Those processes only handle the querying part and some comparison. This one also ensures we process the cycles so that we can resume the normal archiver functioning

aniketdivekar (approved):

PR #149, opened 2025-02-28
SHARD-1875: collectCycleData: only consider incoming cycle for processing

ahmxdiqbal:


urnotsam (dismissed):

kun6fup4nd4 (dismissed):

urnotsam (approved):

src/Data/Data.ts
```diff
@@ -357,14 +357,18 @@ export function collectCycleData(
 ): void {
   for (const cycle of cycleData) {
     // Logger.mainLogger.debug('Cycle received', cycle.counter, senderInfo)
+    let cycleToSave = []
+
+    if (NodeList.activeListByIdSorted.length > 0) {
+      const [ip, port] = senderInfo.split(':')
+      const isInActiveNodes = NodeList.activeListByIdSorted.some(node => node.ip === ip && node.port.toString() === port)
+      const isInActiveArchivers = State.activeArchivers.some(archiver => archiver.ip === ip && archiver.port.toString() === port)
+      if (!isInActiveNodes && !isInActiveArchivers) break
```
aniketdivekar, line 366: - why are we breaking the loop? shouldn't we continue to next cycle?

- nitpick: this logic is independent of of cycle obj... can be done only once right before the loop(unless `NodeList.activeListByIdSorted` or `State.activeArchivers` are updated in this loop.)

- nitpick: we added `NodeList.activeListByIdSorted.length > 0` but why no `State.activeArchivers.length` check?

(I don't have full understanding of this flow)

ahmxdiqbal: 1. if we dont trust the sender, there is no point iterating over the rest of the elements in the list, since the list is input.
2. Youre right, the logic would be the same, but it makes more sense to move it out of the loop
3. We only added the Nodelist check since it was needed to get the network to start. We could also add the archiver length check, but i didnt do it since it wasnt needed

ahmxdiqbal: looking at the code again, for point 2, i think we can keep the sender info check where it is. if the info is invalid, we'll break out of the loop anyways. we wont even have to make the check a 2nd time. in any case, its just a refactor, lmk if you want to do it

aniketdivekar (approved):

PR #152, opened 2025-03-07
SHARD-1923: Use cert scores to determine which cycle received from subscribed nodes to save 

ahmxdiqbal:


src/Data/Cycles.ts
```diff
@@ -267,7 +268,7 @@ function updateNodeList(cycle: P2PTypes.CycleCreatorTypes.CycleData): void {
 
   NodeList.addNodes(NodeList.NodeStatus.SYNCING, consensorInfos)
 
-  NodeList.setStatus(NodeList.NodeStatus.ACTIVE, activatedPublicKeys)
+  NodeList.setStatus(NodeList.NodeStatus.ACTIVE, activatedPublicKeys);
```
aniketdivekar: nitpick: prettier removed `;` if I am correct
```suggestion
  NodeList.setStatus(NodeList.NodeStatus.ACTIVE, activatedPublicKeys)
```

src/Data/Data.ts
```diff
@@ -758,10 +779,10 @@ export async function subscribeNodeFromThisSubset(nodeList: NodeList.ConsensusNo
   }
   if (subscribedNodesFromThisSubset.length > numberOfNodesToSubsribe) {
     // If there is more than one subscribed node from this subset, unsubscribe the extra ones
-    for (const publicKey of subscribedNodesFromThisSubset.splice(numberOfNodesToSubsribe)) {
-      Logger.mainLogger.debug('Unsubscribing extra node from this subset', publicKey)
-      unsubscribeDataSender(publicKey)
-    }
+    // for (const publicKey of subscribedNodesFromThisSubset.splice(numberOfNodesToSubsribe)) {
```
urnotsam: please remove dead code

src/Data/Data.ts
```diff
@@ -352,104 +354,123 @@ export function initSocketClient(node: NodeList.ConsensusNodeInfo): void {
 }
 
 export function collectCycleData(
-  cycleData: P2PTypes.CycleCreatorTypes.CycleData[],
+  cycleData: any[],
```
urnotsam: If the CycleData no longer is applicable can we create a new typing to support the new cycleData structure

src/Data/Data.ts
```diff
@@ -2237,3 +2258,63 @@ async function downloadOldCycles(
...
+    const cleanCert: P2PTypes.CycleCreatorTypes.CycleCert = {
+      marker: cert.marker,
+      sign: cert.sign,
+    }
+    // this has to be disabled until nodelist hashes are set in Q3 in core
+    if (cleanCert.marker !== inpMarker) {
+      Logger.mainLogger.warn('validateCerts: cleanCert.marker did not match inpMarker')
+      return false
+    }
+    if (NodeList.activeListByIdSorted.some(node => node.publicKey === cleanCert.sign.owner) === false) {
```
aniketdivekar, line 2296: is the comparison of output of .some to false correct?

src/Data/Data.ts
```diff
@@ -2237,3 +2258,63 @@ async function downloadOldCycles(
...
+  //console.log('certs:', certs)
+  // look into sign malleability
+  for (const cert of certs) {
+    if (certSigners.has(cert.sign.owner)) continue
+    const cleanCert: P2PTypes.CycleCreatorTypes.CycleCert = {
+      marker: cert.marker,
+      sign: cert.sign,
+    }
+    // this has to be disabled until nodelist hashes are set in Q3 in core
+    if (cleanCert.marker !== inpMarker) {
```
aniketdivekar, line 2292: we are just returning false from here.. what's the purpose? 

as I see the flow =>
* there is a set of markers
* we check if owner is already part of it, if not we continue to next
* we now create a new object and then compare its marker to input marker, 
* if it doesn't match we simply return, but shouldn't we be checking for rest of the certs to check if any of them matching with inpMarker? or do we require that all cert.marker should match inpMarker? 

would be great to have some write up on top of the method telling purpose & logic of the method

src/Data/Data.ts
ahmxdiqbal: yes, since we want to return if the signer is not an active node

src/Data/Data.ts
ahmxdiqbal: the certs are supposed to be the top certs for that marker. if the marker in the cert doesn't contain the same mark as the marker, then it is not a valid cert for this marker. The situation where an honest node will have certs were the markers of its certs are different doesn't seem possible, so i don't see the point in checking if the other certs have valid values. I don't want to carve out an exception for a case only a malicious node would trigger. I could do as you say and check other certs in case one of their markers doesnt match in secure way, but i cant think of a reason to

src/Data/Data.ts
afostr: I think we need to put that check behind an archiver config flag due to how the restore process works a bit differently.

afostr: in restore , the seed note may not go active for like 50 cycles

src/Data/Data.ts
```diff
@@ -759,10 +780,10 @@ export async function subscribeNodeFromThisSubset(nodeList: NodeList.ConsensusNo
   }
   if (subscribedNodesFromThisSubset.length > numberOfNodesToSubsribe) {
     // If there is more than one subscribed node from this subset, unsubscribe the extra ones
-    for (const publicKey of subscribedNodesFromThisSubset.splice(numberOfNodesToSubsribe)) {
-      Logger.mainLogger.debug('Unsubscribing extra node from this subset', publicKey)
-      unsubscribeDataSender(publicKey)
-    }
+    // for (const publicKey of subscribedNodesFromThisSubset.splice(numberOfNodesToSubsribe)) {
```
urnotsam: im not sure why this is commented out.  doesnt seem to have anything to do with cert score changes

src/Data/Data.ts
ahmxdiqbal: done

src/Data/Data.ts
ahmxdiqbal: i wasa dev change, undid it

src/Data/Data.ts
ahmxdiqbal: added the comments btw

urnotsam (dismissed):

aniketdivekar (dismissed):

src/Data/Data.ts
```diff
@@ -352,104 +354,123 @@
...
-      if (value['receivedTimes'] >= minCycleConfirmations) {
-        cycleToSave.push(receivedCycleTracker[cycle.counter][cycle.marker].cycleInfo)
-        value['saved'] = true
       }
-    }
 
-    if (cycleToSave.length > 0) {
-      processCycles(cycleToSave)
+      processCycles([receivedCycleTracker[cycle.counter][bestMarker].cycleInfo])
+      receivedCycleTracker[cycle.counter]['saved'] = true
```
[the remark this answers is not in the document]
mssabr01: This looks valid to me. Please check that cycle.counter is a number before using it to index into receivedCycleTracker, or show where it is previous validated. I don't see ajv between receiving this data and this line yet

src/Config.ts
```diff
@@ -249,6 +250,7 @@ let config: Config = {
   maxRecordsPerRequest: 200,
   multisigKeysSyncFromNetworkInternal: 600,
   minCycleConfirmationsToSave: -1,
+  nerfNonFoundationCertScores: false,
```
mssabr01: Should this be true? Default config should be what we are looking to run on mainnet

mssabr01 (changes requested):

src/Config.ts
urnotsam: its currently set to false on server side.  these should match.  Will get confirmation on what this should be

src/Data/Data.ts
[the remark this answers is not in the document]
urnotsam: if (!validateCycleData(cycle))

we validate that its a number here

mssabr01 (approved):

urnotsam (approved):

aniketdivekar (approved):

PR #245, opened 2025-05-09
SHARD-2423 - feat: add duplicateReceiptsCheckActivationCycle config and enforce ac…

BelfordZ:
…tivation cycle guard in storeReceiptData

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -85,8 +85,6 @@ export interface Config {
   }
   newPOQReceipt: boolean
   storeReceiptBeforeStates: boolean
-  // cycle number from which to activate storeReceiptBeforeStates behavior
-  storeReceiptBeforeStatesActivationCycle: number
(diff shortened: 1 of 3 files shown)


src/Config.ts
```diff
@@ -287,6 +289,8 @@ let config: Config = {
   formingNetworkCycleThreshold: 30,
   maxResponseSize: 15 * 1024 * 1024, // 15MB
   enableDuplicateReceiptsCheck: true,
+  // cycle threshold for duplicate receipts overwrite check
+  duplicateReceiptsCheckActivationCycle: 0,
```
BelfordZ, line 293: We need to set this in env-specific configs. Mainnet / stagenet should be set based on the current cylce number (adding a buffer to acccount for deployment time)

urnotsam (approved):

mhanson-github (approved):

mgthuramoemyint (approved):

PR #264, opened 2025-05-26
SHARD-2463: Handle receipt and original tx thorugh storeReceiptData

S0naliThakur:


src/Data/Collector.ts
```diff
@@ -728,7 +728,8 @@ export const storeReceiptData = async (
   receipts: Receipt.Receipt[] | Receipt.ArchiverReceipt[],
   senderInfo = '',
   verifyData = false,
-  saveOnlyGossipData = false
+  saveOnlyGossipData = false,
+  checkpoint: boolean = true
```
aniketdivekar, line 732: this should be by default false right? only from checkpointing code we should pass it as true.

src/Data/Collector.ts
```diff
@@ -1297,7 +1298,8 @@ export const storeAccountData = async (restoreData: StoreAccountParam = {}): Pro
 export const storeOriginalTxData = async (
   originalTxsData: OriginalTxsData.OriginalTxData[] = [],
   senderInfo = '',
-  saveOnlyGossipData = false
+  saveOnlyGossipData = false,
+  checkpoint: boolean = true
```
aniketdivekar: we don't need this change as no where are we invoking storeOriginalTxData - its technically deprecated

src/Data/Collector.ts
S0naliThakur: This ensures consistency when creating new checkpoints as data is inserted into the database. By default, the database methods set this flag to true. Only the checkpoint system sets it to false, which is why it's true by default here.

src/Data/Collector.ts
S0naliThakur: You are right, it's deprecated and not used, but we made the change for consistency, just to keep things aligned across the board. That said, I can remove it if you prefer.

aniketdivekar: Please go ahead and revert the changes. If need be, add a note on top of the method that its deprecated.

aniketdivekar (dismissed):

urnotsam (dismissed):

jintukumardas (dismissed):

src/Data/Collector.ts
```diff
@@ -765,8 +766,9 @@ export const storeReceiptData = async (
         continue
       }
       if (
-        (processedReceiptsMap.has(txId) && processedReceiptsMap.get(txId) === timestamp) ||
-        (receiptsInValidationMap.has(txId) && receiptsInValidationMap.get(txId) === timestamp)
+        checkpoint &&
```
devendra-shardeum, line 769: please do not add the checkpoint flag here.. since the invalidation maps needs to be checked regardless otherwise it'll skip and do the processing of receipts which has already been seen in past

src/Data/Collector.ts
```diff
@@ -788,7 +790,7 @@ export const storeReceiptData = async (
       if (config.enableDuplicateReceiptsCheck && !receipt.globalModification) {
         // only consider this for EVM txns and Non Global Internal Txns
         const result = await checkIfValidOverwrite(receipt, txId)
-        if (!result) {
+        if (!result && checkpoint) {
```
devendra-shardeum, line 793: this should always be checked regardless of checkpoint enabled or not.. why did we add this flag here ?

src/checkpoint/ReceiptData.ts
```diff
@@ -115,7 +116,7 @@ async function updateData(data: CheckpointData<ReceiptType>): Promise<void> {
     // Insert/Update into checkpoint_data table
     const receipt = data.d
     // Avg entry size is about 41000 bytes
-    await insertReceipt(receipt, false)
+    await storeReceiptData([receipt], 'checkpoint', false, false, false)
```
devendra-shardeum, line 151: Please update all calls to storeReceiptData to explicitly pass the flags as arguments. This will improve code readability and make it clearer what each call is intending to do.

image

devendra-shardeum (changes requested):

src/Data/Collector.ts
devendra-shardeum: also this way we protect our archiver from potential dos as well.

S0naliThakur: The checks will always happen in case of normal flow. But since checkpoint works for recent transactions only, this check will not let us save the majority agreed upon receipt that's why checkpoint flag is needed here.

src/Data/Collector.ts
S0naliThakur: As addressed in the slack conversation, this is needed to save the newly agreed upon success receipt.

src/checkpoint/ReceiptData.ts
S0naliThakur: Added

devendra-shardeum (dismissed):

urnotsam (dismissed):

urnotsam:
Please fix failing tests

urnotsam (approved):

mhanson-github (approved):

